PR #4, opened 2021-12-25
fix duration display

felixguendling:
see https://github.com/valhalla/valhalla/discussions/3474


TimMcCauley:
Thanks @felixguendling keep them coming 👍🏽 👍🏽 👍🏽

PR #19, opened 2022-03-28
car = auto for locate

nilsnolde:
missed one spot to update the costing model to auto instead of car


nilsnolde:
Will never understand why it’s auto😅

PR #36, opened 2022-11-19
use `via` waypoints for intermediates

### LINKED ISSUE #29: Set „type=via“ on intermediate waypoints
Ref https://github.com/valhalla/valhalla/issues/3680

as long as we don’t allow to set a specific type for each waypoint in the app, we should set an explicit „type“ of „via“. Would actually be nice to have a dropdown to select the type eventually.

EDIT: might also be more issues in how we parse the response. I just repeated what Dave said in the Valhalla issue, didn’t look into it at all.

nilsnolde:
fixes #29


nilsnolde:
well, you're right though. for readability I'll sacrifice those 2 mins:)

Issue #31, opened 2022-09-28
Truck's default width & height are describing car's defaults


nilsnolde:
Just had a confusion about the costing defaults of the web app at https://github.com/valhalla/valhalla/discussions/3767. I realized we're swapping the defaults Valhalla uses for `width` & `height`. Quick fix.

nilsnolde:
Sorry, had the wrong request. In fact, our network analyst has that similar problem :sweat_smile: 

The web app also has a problem though: it's assuming a width of 1.6 m and a height of 1.9 m for truck, while that configuration is the default for cars.

PR #37, opened 2022-11-19
set truck width & height properly

nilsnolde:
fixes #31

it's not the prettiest but gets the job done;)


src/Controls/Directions/index.jsx
```diff
@@ -30,6 +31,7 @@ class DirectionsControl extends React.Component {
   handleUpdateProfile = (event, data) => {
     const { dispatch } = this.props
     dispatch(updateProfile({ profile: data.valhalla_profile }))
+    dispatch(resetSettings())
```
nilsnolde, line 34: reset settings when profile updates

src/Controls/settings-options.js
```diff
@@ -520,6 +520,12 @@ export const settingsInit = {
   use_geocoding: true,
 }
 
+export const settingsInitTruckOverride = {
+  ...settingsInit,
+  width: 2.6,
+  height: 4.11,
+}
+
```
nilsnolde, line 528: urgh I know, but wasn't feeling like more of a rewrite

src/Controls/index.jsx
```diff
@@ -124,6 +125,7 @@ class MainControl extends React.Component {
         }
       })
       dispatch(zoomTo(processedCoords))
+      dispatch(resetSettings())
```
nilsnolde, line 128: this will reset when it's loaded from a URL

chrstnbwnkl (approved):
fine for me, at this point it's gotta be either a dirty quick fix or an entire refactoring, nothing in between :smile:

nilsnolde:
agreed!

Issue #15, opened 2022-03-22
Responsive layout for small (phone) screens


westnordost:
The app's layout is a bit of a mess on phone screens, see:

![image](https://user-images.githubusercontent.com/4661658/159525694-b653eba6-bcf9-470b-b418-0f5d08b930f8.png)

Notes / Possible solutions / Thoughts:
- https://maps.openrouteservice.org/#/ does the responsive-ness on small screens quite well but probably very different to this app. All the advanced features and options are hidden in a sidebar
- If the width is below a certain threshold, the floating-dialog-like box could become full-width (without margin and shadow), this saves a few pixels
- move the small buttons (delete, add, settings) to the bottom (and/or own row). This makes the whole dialog less broad
- just use the common "drag handle" icon for the drag handle, i.e. no numbers, not styled as a button. I think would also be better usability-wise, because it simply is no button and the concept+icon of drag handles is reasonably well understood E.g. like here https://magento.stackexchange.com/questions/272519/how-to-add-draggable-handle-for-configure-product-variants-in-magento-2
- if there is (still) not enough space, it is a common pattern to ellipsize the button bar at the top, the remaining transport modes being behind a "..." button (in a dropdown menu)
- the settings dialog should respect the boundaries of the main dialog, i.e. not appear behind it. If there is not enough space horizontally, it could go below the main dialog. It may be worth a thought to not show the settings dialog by default and/or only if the screen is wide enough

This feature request was triggered by https://github.com/streetcomplete/StreetComplete/issues/3905 , a request to add the valhalla demo to the list of useful OpenStreetMap showcase links in StreetComplete.

TimMcCauley:
Hi @westnordost - thank you for raising this. Indeed, this should be done. Most of it can be solved with a bit of css. Let me dive in!

georg-d:
While [#18](https://github.com/gis-ops/valhalla-app/issues/18) is also about the mobile experience, it's not an exact dupe but rather about which details might be improved - so maybe it's helpful to solve [#15](https://github.com/gis-ops/valhalla-app/issues/15)

TimMcCauley:
We have been brainstorming a little about this. Going forward with this will take some time which we currently do not have. I found this library https://mui.com/material-ui/react-drawer/ which we now use for the range sliders, too. The drawers look pretty decent and they could help us move towards a responsive layout - thoughts? Is there anything you could and would like to pick up here @georg-d  @westnordost ?

westnordost:
[...]

PR #39, opened 2022-11-19
Improve mobile layout

wcedmisten:
Closes https://github.com/gis-ops/valhalla-app/issues/15

Improves the layout to be a bit more mobile friendly.

Minor regression: the drawer takes up the entire vertical space, which obscures part of the map on desktop. I tried fixing this by adding `height: undefined` to the style, but this broke the ability to scroll when the content expands past the screen.

# Changes
* use drawer for "Directions/Isochrones" modal
* use drawer for the settings modal
* don't show the settings page by default
* adds close button for the Directions/Isochrones drawer, allowing user to see the whole map
* adds a button to open the Directions/Isochrones drawer from the map view

You can also view a demo of this (only on North America) at https://map-demo.wcedmisten.dev, if you want to check it out on a true mobile device.

# Desktop Layout
image
image
image

# Mobile layout
image
image
image

Sorry the commit history is a bit ugly because I had to revert the changes from my fork.


nilsnolde:
Wow, that's great! Thanks so much @wcedmisten ! We didn't have much time lately to care about this repo. I think this is a great start! I tried it on an iPhone 11 Pro and it looks really good. On Desktop it could be better I guess, but I'd argue it was worse before, now at least it's hideable sidebars. 

There's one more smaller thing I'd do (nothing related to your PR):
image

The button group on the upper right: I think it's a weird arrangement, the settings button should be more prominent. IMO the settings button should be as big as the profile buttons and the "reset all" & "add via" buttons could be in the row below all locations. Then there's a bit more space in that row. Also, the settings button could be highlighted somehow, as that's hiding the whole meat of Valhalla.

What do you think @wcedmisten ? No problem if you'd like to skip that, can always do it in a later PR. Honestly this is great. We're about to apply for a renewal of the funding for the servers, so it's a great timing:)

src/actions/directionsActions.js
```diff
@@ -81,10 +81,10 @@ const fetchValhallaDirections = (valhallaRequest) => (dispatch) => {
       'Content-Type': 'application/json',
     },
   }
-
   axios
     .get(VALHALLA_OSM_URL + '/route', config)
     .then(({ data }) => {
+      console.log(data)
```
nilsnolde, line 87: debug statement

wcedmisten:
Thanks for the feedback! I went ahead and implemented the changes you suggested. It looked a little weird just having the reset/add buttons alone in the next row, so I moved them to the bottom for Directions and next to the search bar for Isochrones.

image
image

nilsnolde (approved):
Great, totally agree, good job with the re-arrangement! Can’t wait to see it on the official site, I’ll deploy later tonight. Thanks a ton!

PR #42, opened 2022-11-21
modified initial heightgraph width

chrstnbwnkl:
We previously did not set an initial width on the height graph, so the svg dimensions were always 800px regardless of device dimensions. This pr addresses that. Caveat: does not change on resize at the moment.


nilsnolde:
LGTM, thanks!

PR #46, opened 2022-11-29
Add Dockerfile

wcedmisten:
From https://mherman.org/blog/dockerizing-a-react-app/#production

Some previous discussion asked how to run this in Docker, so I thought adding it made sense. It's a fairly simple Dockerfile.

https://github.com/gis-ops/valhalla-app/discussions/24

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,18 +5,18 @@ yarn.lock
 .idea/
 
 # dependencies
-/node_modules
-/.pnp
+node_modules
+.pnp
 .pnp.js
 
 # testing
-/coverage
+coverage
 
 # production
-/build
+build
 
 # misc
-.DS_Store
+**/.DS_Store
 .env.local
 .env.development.local
 .env.test.local
@@ -28,7 +28,6 @@ yarn-error.log*
 
 # docker
 .dockerignore
-node_modules
 npm-debug.log
 Dockerfile
 .git
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,36 @@
+package-lock.json
+yarn.lock
+
+# IDEs
+.idea/
+
+# dependencies
+/node_modules
+/.pnp
+.pnp.js
+
+# testing
+/coverage
+
+# production
+/build
+
+# misc
+.DS_Store
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
+
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+# docker
+.dockerignore
+node_modules
+npm-debug.log
+Dockerfile
+.git
+.gitignore
+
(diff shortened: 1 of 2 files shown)


ImreSamu:
@wcedmisten : 
Thank you! 

Would it be possible to use a "Supported tags" from https://hub.docker.com/_/node
now it is too old:
```bash
docker inspect  node:13.12.0-alpine | grep Created
        "Created": "2020-03-28T00:25:41.102584861Z",
```

other security advice: https://snyk.io/blog/10-best-practices-to-containerize-nodejs-web-applications-with-docker/
-  `.dockerignore` 
 -  `USER node`  + `COPY --chown=node:node ...`

nilsnolde:
Thanks @wcedmisten!

Looks good, but I agree with @ImreSamu:
- a more recent version of the build image would be cool (16.x?)
- a .dockerignore file is mandatory, mostly the same as .gitignore
- the rest, USER etc I don’t mind, there’s nothing exposed but a port


Btw, you can clone directly from this repo now and create branches.

wcedmisten:
Thanks for the feedback, that was a good article on securing a node Dockerfile.

I updated the node base image version and added a `.dockerignore` file.

I tried setting a non-root `USER`, but it seems like the nginx image would require some additional configuration to accommodate that, so I left it as the simpler root user for now.

.dockerignore
```diff
@@ -0,0 +1,36 @@
+package-lock.json
+yarn.lock
+
+# IDEs
+.idea/
+
+# dependencies
+/node_modules
```
nilsnolde, line 8: Probably fine, but you sure this is interpreted as relative path and not absolute?

nilsnolde: I see it's taken from the .gitignore. Same question there:) But also pretty sure .dockerignore & .gitignore are not handled the exact same, so behavior might differ.

.dockerignore
```diff
@@ -0,0 +1,36 @@
...
+.env.test.local
+.env.production.local
+
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+# docker
+.dockerignore
+node_modules
```
nilsnolde, line 31: this is more what I'm used to

.dockerignore
wcedmisten: Yeah, it looks like with `.dockerignore` the paths are interpreted as only under the root. So here `/node_modules` is the same as `node_modules` in the other line. I think most of these would only appear under the root anyway, so that's fine.

I updated `.DS_Store` to `**/.DS_Store` because that could appear under any directory for MacOS users.

.dockerignore
wcedmisten: removed the duplicate `node_modules`

nilsnolde (approved):

PR #49, opened 2023-01-05
fixed zooming on route update

### LINKED ISSUE #48: Prevent zooming out on short routes
When calculating routes between points that are not far from each other, the map view always zooms out, so that the route becomes way to small to see.

chrstnbwnkl:
fixes #48

### DIFF
--- a/src/Map/Map.jsx
+++ b/src/Map/Map.jsx
@@ -415,7 +415,7 @@ class Map extends React.Component {
 
   zoomToCoordinates = () => {
     const { coordinates } = this.props
-    this.map.fitBounds(coordinates, { padding: [50, 50], maxZoom: 11 })
+    this.map.fitBounds(coordinates, { padding: [50, 50] })
   }
 
   zoomTo = (idx) => {
--- a/src/Map/Map.jsx
+++ b/src/Map/Map.jsx
@@ -116,6 +116,8 @@ class Map extends React.Component {
     activeDataset: PropTypes.string,
     showRestrictions: PropTypes.object,
     coordinates: PropTypes.array,
+    showDirectionsPanel: PropTypes.bool,
+    showSettings: PropTypes.bool,
   }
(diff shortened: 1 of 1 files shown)


chrstnbwnkl:
> So now it’ll automatically zoom in to new routes & isochrones, no matter how small, with a 50 px padding?

There is no zooming on isochrones yet, but I could look into that as well. But yes, it zooms into routes no matter how short.


> That function isn’t really used for anything else like the geocoding results or so?

Well now we know why that `maxZoom` was there in the first place :smile: The problem is, there doesn't seem to be an easy way to distinguish between a route zoom action and a geocode zoom action, other than the coordinate array length (always 1 for geocoding).  Since a route will always have at least two points (right?), we could do this instead of passing information from the caller.

One last thing I noticed is that since the drawer sits on top of the map container, the route will not be properly zoomed into view when the drawer is open, but there seems nothing we can do about it unless we pass the drawer state to the map (at least it has a fixed pixel size so we'd be able to calculate left padding based on that).

nilsnolde:
> Since a route will always have at least two points (right?)

Right:)

> the route will not be properly zoomed into view when the drawer is open

Hm, what is "the drawer"?

chrstnbwnkl:
Sorry, by drawer I mean the collapsible side bar (commonly referred to as a [drawer](https://cultureamp.design/components/drawer/), also the React component we use is called that way :smile:).

nilsnolde:
Right, thanks. If we could respect the drawer it'd be nicer, but on mobile that'd be a problem. If it's easy enough to avoid that, let's do it. If not, then I'd also just take the entire width (for now). What do you think?

chrstnbwnkl:
Was easy enough, so I also took the settings panel on the right side of the screen into account, starting at a certain screen width.

nilsnolde (approved):

Issue #12, opened 2022-03-11
time is reported wrong


nilsnolde:
reported first by @georg-d in https://github.com/valhalla/valhalla/issues/3573

e.g. https://valhalla.openstreetmap.de/directions?profile=pedestrian&wps=9.884691238403322%2C47.599319584622%2C9.872074127197267%2C47.602632914527724

reports 1:15 h where valhalla only responds with 936 secs = 15.6 mins. somehow there seems to be 1 hour added.

PR #51, opened 2023-01-16
Update route summary tooltip with correct time

wdcsimon:
Time is reported wrong in route summary tooltip when hovering over route.  Apply same fix as used in Issue #12

### DIFF
--- a/src/Map/Map.jsx
+++ b/src/Map/Map.jsx
@@ -15,6 +15,7 @@ import PropTypes from 'prop-types'
 import axios from 'axios'
 
 import * as R from 'ramda'
+import { intervalToDuration } from 'date-fns'
 import ExtraMarkers from './extraMarkers'
 import { Button, Label, Icon, Popup } from 'semantic-ui-react'
 import { CopyToClipboard } from 'react-copy-to-clipboard'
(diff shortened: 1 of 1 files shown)


src/actions/commonActions.js
```diff
@@ -63,7 +63,7 @@ export const updatePermalink = () => (dispatch, getState) => {
   const queryParams = new URLSearchParams()
   queryParams.set('profile', profile)
 
-  let path = '/directions?'
+  let path = '#directions?'
```
nilsnolde, line 66: This seems wrong, can you elaborate here?

nilsnolde (commented):
Thanks for the fix. Just one thing to clarify.

src/actions/commonActions.js
wdcsimon: Basically "directions" or "isochrones" are not physical files, but handlers, so they should follow a # rather that a / as this provides correct URL routing.  If we use the / then if a user does a browser page refresh, they end up generating a 404-error page, by using the # we avoid this potential issue.  See [https://maps.whanganui.govt.nz/routing](https://maps.whanganui.govt.nz/routing) to see this in action.

nilsnolde: > Basically "directions" or "isochrones" are not physical files, but handlers

Right, as a web dev noob I figured that's mostly syntactical in real life if using `/` or `#`.

What I can see on your demo page, it's not properly resolving permalinks, i.e. page refreshes fail to parse the full URL into a parameterized version of the web app. E.g. this: https://maps.whanganui.govt.nz/routing/#directions?profile=bicycle&wps=175.02682766786324%2C-39.91050190801741%2C175.04880524660373%2C-39.93498734998717, would generate a route from the waypoints on https://valhalla.openstreetmap.de/.

Wouldn't you also have to update the function which parses the URL from a permalink?

nilsnolde: Though I'm still open to adhering to proper standards (if that's what it is), but another point is people already pointing to the app, in all the issues in upstream Valhalla etc. If there's no really good reason to change it, I wouldn't like to include that in the merge, it'd mess up all existing links (I know I can tell Apache to deal with this, but again, needs a good reason for me).

wdcsimon: Good point about the permalink, the # doesn't help with that. Will look at how that can be resolved.
Probably better to stick with the / at the moment.

nilsnolde: I think @chrstnbwnkl convinced me about the `#`, I always wondered why so many SPAs do that. Anyways, I guess we'll just need to update the URL parsing function to trigger permalinks on `#`. I'd guess it could be done in a backwards-compatible way so that old `/` URLs are still resolved properly, e.g. from open Valhalla issues. There's custom code in the app for that, I remember. 

If you don't have the time for that @wdcsimon let us know and we'll handle it at some point. Thanks anyway!

PR #62, opened 2023-02-20
Route summary reporting error fixed

rakeshhotker:
changed the code for formatDuration function defined in Summary.jsx to be same as the one defined in Map.jsx. As the definition in Map.jsx seemed more appropriate. Do let me know if any modifications are required.
Update: Also did the enhancement as required of showing the user the tileset age. Do let me know if any modifications are required.

### DIFF
--- a/src/Controls/Directions/Summary.jsx
+++ b/src/Controls/Directions/Summary.jsx
@@ -2,7 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
 import * as R from 'ramda'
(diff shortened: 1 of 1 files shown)


nilsnolde:
You did not mention this is for GSoC, see #67 . We can open it again once we get to this point.

nilsnolde:
Now that we have some guidelines, I'll open this again. https://github.com/gis-ops/valhalla-app/wiki/GSoC-2023

nilsnolde:
I'd just ask you to revert the tileset age thing and focus on just one issue.

PR #69, opened 2023-02-27
Route summary reporting error fixed

rakeshhotker:
As told, i reverted to the Route summary reporting error so as to focus on one bug.

### DIFF
--- a/src/Controls/Directions/Summary.jsx
+++ b/src/Controls/Directions/Summary.jsx
@@ -2,7 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
 import * as R from 'ramda'
-import { intervalToDuration } from 'date-fns'
+//import { intervalToDuration } from 'date-fns'
 
 import { Icon, Checkbox, Popup } from 'semantic-ui-react'
 import { showProvider } from '../../actions/directionsActions'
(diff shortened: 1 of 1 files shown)


src/Controls/Directions/Summary.jsx
```diff
@@ -22,38 +21,22 @@ class Summary extends React.Component {
...
-    }
-    if (duration.seconds > 0) {
-      durationStr += duration.seconds + 'sec'
-    }
-    return durationStr
+    const date = new Date(durationInSeconds * 1000)
+    const days = date.getDate() - 1 > 0 ? date.getDate() - 1 + 'd ' : ''
+    const hours = date.getHours() > 0 ? date.getHours() + 'h ' : ''
+    const minutes = date.getMinutes() > 0 ? date.getMinutes() + 'min' : ''
+    return days + hours + minutes
```
nilsnolde, line 28: not DRY, please outsource and re-use

nilsnolde (changes requested):

src/Controls/Directions/Summary.jsx
nilsnolde: although, now there's #91 , so likely the other place where this is done, will change to a hh:mm:ss format. maybe we should do that here too?

nilsnolde: ended up not using hh:mm:ss format, but compress it a bit regardless in #94.  we should re-use that function.

rakeshhotker:
Could you please be more descriptive

nilsnolde:
You copy/pasted this function as-is from another place in the same codebase. You're repeating the same stuff, that's never good.

Did you look at #94? You'll need to merge master, then you can apply your changes while having the function live somewhere else and called by both modules.

aaryahjolia:
Can I make a PR on this @nilsnolde ? I solved this issue by DRY principle and in hh:mm:ss format.

nilsnolde:
Depends if @rakeshhotker is dropping the contribution or not, I'd still give him some time.

Issue #50, opened 2023-01-05
Remember map extent


nilsnolde:
We should do the same as most map apps and remember the last extent when visiting the website. It's safe to just put it in the local browser store IMO. Smth we might want to do for more info than just extent, e.g. last used costing model?

aa98-45556443355666:
I want to work on this issue for GSOC 2023. Kindly assign to me.

nilsnolde:
Sure, once you put a PR, we can review.

aa98-45556443355666:
ok I got it.

hs7898753:
Hi, I want to work on this issue please assign it to me

nilsnolde:
@aa98-45556443355666 we're open now, feel free to start and follow our guidelines: https://github.com/gis-ops/valhalla-app/wiki/GSoC-2023

aa98-45556443355666:
#87

aa98-45556443355666:
I will add a fresh pull request which will solve this problem also. [https://github.com/gis-ops/valhalla-app/issues/85](url)

nilsnolde:
Let’s give others a chance

aa98-45556443355666:
Please assign this issue to  me.

inceptionabhishek:
@nilsnolde I want to take this issue. I have already raised a similar issue recently. #85

aa98-45556443355666:
**@nilsnolde**  Its been more than 4 days that nobody has assigned this issue. Its  my humble request to assign this issue to me as I have   implemented the feature to remember map extent.

nilsnolde:
I tried telling you in the previous PR, where you just sold someone else's work as your own, that it's not the way. We already have tons of applicants who keep me quite busy (compared to other projects I'm mentoring), so we can be a bit stricter here and I can tell you that we'll go with someone else. It's not like people don't deserve second chances, but also there's other GSoC projects, so I suggest you focus on smth else and just take this as a lesson.

PR #73, opened 2023-02-28
added remember map extent (#50)

sandstone991:
Hello, sorry if this is against rules because I didn't ask to be assigned first.
Anyway added a simple debounce to store the last lat & lng the user viewed as well as the zoom level, and in the componentDidMount life cycles I check for these items in the localStorage if they're there I take them and replace the default mapParams.

### DIFF
--- a/src/Map/Map.jsx
+++ b/src/Map/Map.jsx
@@ -30,6 +30,7 @@ import {
   buildLocateRequest,
 } from 'utils/valhalla'
 import { colorMappings, buildHeightgraphData } from 'utils/heightgraph'
+import { throttle } from 'throttle-debounce'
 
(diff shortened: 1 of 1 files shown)


nilsnolde:
I think this is fine technically, but our bigger point, apart from communicating that you want to work an issue for GSoC, was that you shouldn’t do **more than 1**.

You’re interested in GSoC I assume right?

sandstone991:
@nilsnolde 

> I think this is fine technically, but our bigger point, apart from communicating that you want to work an issue for GSoC, was that you shouldn’t do **more than 1**.

Yeah I misunderstood that, sorry for the inconvenience though.

> You’re interested in GSoC I assume right?
 
Yes.

nilsnolde:
closing for now and will merge your other pr

PR #79, opened 2023-03-03
fix: results vertical ordering w.r.t active tab

### LINKED ISSUE #78: fix: showing results for respective tab
**Problem**
Right now even if we have switched to the isochrones tab, we are still showing results of the directions tab at the bottom. After entering name of a place for which we want to generate an isochrone we don't know that the results are at the bottom of the sidebar below the results of directions. This is not very convenient for users.

**Solution**
We can show the results depending upon the respective tab that has been chosen.     

**Screenshot**
https://user-images.githubusercontent.com/55504616/222498026-b86acf99-0f28-4403-99b5-9ac2c6541818.mp4

**Additional note**
I can work on this.

Ananya2001-an:
**Fixes issue**
closes #78 

With respect to the active tab we are showing results at the bottom in the respective order.

**Screenshot**
https://www.awesomescreenshot.com/video/15283492?key=f6d41803f4749982e1d2c825d8131e03

### DIFF
--- a/src/Controls/index.jsx
+++ b/src/Controls/index.jsx
@@ -219,6 +219,7 @@ class MainControl extends React.Component {
                 <ServiceTabs />
               </div>
             </Segment>
+            {/* because apparently on small screens it's not showing both, so we switch the order on tab switch */}
             {(activeTab === 0 && (
               <>
                 <DirectionOutputControl />
--- a/src/Controls/index.jsx
+++ b/src/Controls/index.jsx
@@ -219,8 +219,17 @@ class MainControl extends React.Component {
                 <ServiceTabs />
               </div>
             </Segment>
-            <DirectionOutputControl />
-            <IsochronesOutputControl />
+            {(activeTab === 0 && (
+              <>
+                <DirectionOutputControl />
+                <IsochronesOutputControl />
+              </>
+            )) || (
+              <>
+                <IsochronesOutputControl />
+                <DirectionOutputControl />
+              </>
+            )}
           </div>
         </Drawer>
         <SemanticToastContainer position="bottom-center" />
--- a/src/Controls/Isochrones/OutputControl.jsx
+++ b/src/Controls/Isochrones/OutputControl.jsx
@@ -36,7 +36,7 @@ class OutputControl extends React.Component {
     return (
       <Segment
         style={{
-          margin: '0 1rem',
+          margin: '0 1rem 10px',
           display: successful ? 'block' : 'none',
         }}
       >


Ananya2001-an:
@nilsnolde kindly review the changes.

nilsnolde (commented):
Looks good I think. I'll review (and try out) all PRs this weekend hopefully.

Ananya2001-an:
> Looks good I think. I'll review (and try out) all PRs this weekend hopefully.

Cool!

src/Controls/index.jsx
```diff
@@ -219,8 +219,17 @@ class MainControl extends React.Component {
                 <ServiceTabs />
               </div>
             </Segment>
-            <DirectionOutputControl />
-            <IsochronesOutputControl />
+            {(activeTab === 0 && (
```
nilsnolde, line 222: should we add a comment why we did this? it does seem weird on the first glance when you switch tabs around, that the order switches. but in the end we do it because apparently on small screens it's not showing both.

Ananya2001-an: Yeah I guess we can add a comment but what exactly should be written? I am not sure what to write in one simple sentence to explain this.

nilsnolde: I think this is ok enough: `because apparently on small screens it's not showing both`, maybe plus `, so we switch the order on tab switch`

Ananya2001-an: Okay I will add this comment soon.

Ananya2001-an:
Hey @nilsnolde I have added the comment. I guess you can merge it now.

src/Controls/index.jsx
```diff
@@ -219,8 +219,18 @@ class MainControl extends React.Component {
                 <ServiceTabs />
               </div>
             </Segment>
-            <DirectionOutputControl />
-            <IsochronesOutputControl />
+            {/* because apparently on small screens it's not showing both, so we switch the order on tab switch */}
```
nilsnolde, line 222: sorry for being nitpicky on this, but: can you use `//` line comments for single comments and `/* */` multiline comments for more elaborate ones? Also, move the comment above the `{` on its own line. I'm surprised `prettier` doesn't complain..

nilsnolde: Oh I see, is this somehow related to being inside the React "HTML" stuff? I see the enclosing `{}` and wonder why they're there. Then the way you did it, might be the only valid way to comment in this situation?

Ananya2001-an: Yeah comments in JSX are supposed to be this way.

Issue #92, opened 2023-03-06
Small caps texts in Directions and Isochrones profiles


aditygrg2:
From a UX perspective, all small caps and big caps are not seen as formal. For the units, it is okay as it is written in all small caps only, but the profiles should have mixed caps (first capital, other small) to improve readability.

<img width="1440" alt="Screenshot 2023-03-06 at 14 21 48" src="https://user-images.githubusercontent.com/98523623/223063514-72fedb65-bab1-4e49-98fb-9259c52c3ea8.png">

I can help with this issue.

nilsnolde:
Hm okay :D Can you merge this with your previous issue and just call it "minor UI improvements"? Only one PR would be appreciated:)

Issue #91, opened 2023-03-06
Text besides time icon, breaks the defined size


aditygrg2:
Check the images - 
<img width="1440" alt="Screenshot 2023-03-06 at 13 56 21" src="https://user-images.githubusercontent.com/98523623/223061373-54a0b71a-cccd-4e78-8c52-937537ded3aa.png">
<img width="1440" alt="Screenshot 2023-03-06 at 13 56 13" src="https://user-images.githubusercontent.com/98523623/223061434-9d014182-ca44-4480-bc16-2f3dfb6ff1b1.png">

The text besides the time icon, is not in a single line, which distorts UI and doesn't look good.

This can be improved by either shortening the text, for example, min can be written as m, and seconds as s or we can reduce the text size, or give it more space, but the latter one will require more refactoring, and the first one is easy to fix and UI will not be affected much.

If needed, I can solve this issue as my pre-gsoc work.

nilsnolde:
Yeah I agree. In fact let's do the hh:mm:ss format.

aditygrg2:
@nilsnolde There are days defined as well. I think it will be confusing if we keep hh:mm::ss format, a d,h,m,s format will look more better.

nilsnolde:
Yeah I think that was the reason why we did it this way as it could potentially wrap around 24 hours into the next day. I'm fine with your proposal too, still better than currently.

aditygrg2:
Great, thank you, will add a PR!

PR #94, opened 2023-03-06
Minor UI Changes

aditygrg2:
• Made mixed caps for profiles and some texts app-wide.
• Updated the summary time section, which was breaking in case of long time.

Solves issue https://github.com/gis-ops/valhalla-app/issues/91 and https://github.com/gis-ops/valhalla-app/issues/92.

### DIFF
--- a/src/Controls/Directions/index.jsx
+++ b/src/Controls/Directions/index.jsx
@@ -81,12 +81,12 @@ class DirectionsControl extends React.Component {
                 ]}
                 loading={loading}
                 popupContent={[
-                  'bicycle',
-                  'pedestrian',
-                  'car',
-                  'truck',
-                  'bus',
-                  'motor scooter',
+                  'Bicycle',
+                  'Pedestrian',
+                  'Car',
+                  'Truck',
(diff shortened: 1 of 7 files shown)


src/Controls/Directions/index.jsx
```diff
@@ -81,12 +81,12 @@ class DirectionsControl extends React.Component {
...
-                  'car',
-                  'truck',
-                  'bus',
-                  'motor scooter',
+                  'Bicycle',
+                  'Pedestrian',
+                  'Car',
+                  'Truck',
+                  'Bus',
+                  'Motor Scooter',
```
nilsnolde, line 89: Now looking at it, I'm not sure this is proper: the name of it is also taken as the costing model input to the Valhalla requests IIRC. Did you try the app locally with some routing/isochrone requests?

nilsnolde: Ah sorry, it should be fine of course, we can always make the request option small letters and that might be done already. Just checked this out in the browser and behaves all fine.

PR #96, opened 2023-03-06
Added Timeout For Colour in Copy clipboard (In Settings Sidebar Tab)

inceptionabhishek:
## 🛠️ Fixes Issue

In the setting tab, when we click on copy to clipboard we can see that its color changes to green which is a good user experience. However, it stays Green throughout the application unless the user leaves the platform or refreshes the page.
Even if we click again on the "copy to clipboard", it stays the same, which is obviously not a good user experience.

## 👨‍💻 Changes proposed

I would like it to automatically change to its original color after 1 second or 2 seconds for a better user experience. So, I am suggesting adding a timeout of 1 second, after 1 second **state: copied** will get changed to false, and so do the color.

## 📄 Note to reviewers

Actually, I noticed this issue just after my previous pull request, otherwise, I would have solved this in that previous PR itself.

## 📷 Screenshots

Before : 

Untitled video - Made with Clipchamp

After changes : 

after

### DIFF
--- a/src/Controls/Directions/Waypoints/index.jsx
+++ b/src/Controls/Directions/Waypoints/index.jsx
@@ -82,7 +82,6 @@ class Waypoints extends Component {
 
   render() {
     const { waypoints } = this.props.directions
-    console.log('waypoints', waypoints)
     return (
       <DragDropContext onDragEnd={this.onDragEnd}>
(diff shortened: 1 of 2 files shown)


nilsnolde (approved):
jep, thanks, that's a good improvement.

Issue #68, opened 2023-02-23
Rewrite using functional components


chrstnbwnkl:
While React class components are not going anywhere any time soon, functional components have become the new norm. A rewrite using functional components and hooks could make the application future proof and might offer a way to simplify the code along the way (especially the React <-> Leaflet interactions, which are all buried inside one big Map component).

inceptionabhishek:
Please Assign me this issue, I can work on this.

aa98-45556443355666:
I want to work on this issue . Kindly assign it to me.

nilsnolde:
This is too big for applicants, it'll be part of the main GSoC project.

theWiseAman:
@nilsnolde Can I work on some files though? I have experience as a ReactJS Developer so I know what is being asked here. So can you assign me this issue?

nilsnolde:
Then you'd handle it pretty efficiently once the project started I think. There are lots of open, unassigned issues, please choose one of those. It's more important for the pre-selection to see how you do things you're not familiar with.

samyabrata-maji:
> While React class components are not going anywhere any time soon, functional components have become the new norm. A rewrite using functional components and hooks could make the application future proof and might offer a way to simplify the code along the way (especially the React <-> Leaflet interactions, which are all buried inside one big Map component).

@chrstnbwnkl This is a pretty big issue. It is better to handle it in parts. Would be nice, if you could provide a list of files that need work first.

nilsnolde:
Again, reading capability is a requirement.

samyabrata-maji:
> Again, reading capability is a requirement.

Can you point out some issue that needs solving?

tommach591:
I submitted a proposal for GSoC 2023 to work on this issue and I was wondering if there are any feedbacks on it.

Ananya2001-an:
Hey @nilsnolde do u think I should start working on this on the side....it will take time but better to start now so that later on other issues can also be resolved and anyways I am free ;)

PR #98, opened 2023-03-06
Improved control tabs in Left side panel

### LINKED ISSUE #88: Same width of Directions and Isochrones
Valhalla_Bug
  

As seen in above picture, the width distribution is uneven when we switch to Isochrones, the box or the left pane shrinks some what and due to that, The **2**nd section of image shows uneven distribution.

Also, In the **1**st section, We can give 50-50 width to directions and isochrones so they will look much good.
Something like this:
image

I want to implement and fix both of this issue as a part of my pre-gsoc work.

aaryahjolia:
## 🛠️ Fixes Issue
Closes #88 

## 👨‍💻 Changes proposed

1. The directions and isochrones button were not fully occupying the width, So I extended them to use full width. (See images)
2. The slider in setting in Isochrones section was having a padding-left but not padding-right due to which it was not looking good. I added padding-right also so it aligns in the center with equidistance from both sides.

## 📄 Note to reviewers

While I was roaming in the code base, I found that most of the part of CSS is not written properly, changing one part affects another part adversely. Can I start writing my proposal around refactoring most of the codebase (mostly CSS and adding some new functionalities also) so that it would be written in accordance of standard practices.

## 📷 Screenshots

1. Before:
image
After:  
image

2. Before: 
image
After:  
image

### DIFF
--- a/src/Controls/Isochrones/Waypoints/index.jsx
+++ b/src/Controls/Isochrones/Waypoints/index.jsx
@@ -247,7 +247,7 @@ class Waypoints extends Component {
             </Accordion.Title>
             <Accordion.Content active={activeIndex === 0}>
               <Form size={'small'}>
-                <div className={'pt3 pl3'}>
+                <div className={'pt3 pl3 pr3'}>
                   <Form.Group inline>
                     <Form.Input
                       width={12}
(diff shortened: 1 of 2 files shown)


nilsnolde:
LGTM, thanks!

> I found that most of the part of CSS is not written properly, changing one part affects another part adversely.

I can believe that! We'd like a re-write of the app, CSS/styling handling could very well be part of it. We'd like to do #68 as well.

aaryahjolia:
Excellent @nilsnolde , Can I start writing my proposal of rewriting the code from Class to function based React, Improving CSS, etc. or do I need some more PR before my proposal?

PR #103, opened 2023-03-08
Fixed Max toast count to 1 in case of current profile exceeds its OSM limit distance.

inceptionabhishek:
## 🛠️ Fixes Issue

So, When we select two points in the map ( say source and Destination ), if the current profile exceeds its OSM limit, it sometimes shows multiple toast error messages for the same thing, stacking one above the other.

For example: If I select India to Australia for Car profile, it will show the max limit exceeded in toast right, but due to one more extra API call makes it show two messages, sometimes once which I think can be annoying for some users. 

## 👨‍💻 Changes proposed

In the Toast Container, we can set its max Count to 1, So, it will only show one popup at once in the screen also automatically in order not skipping any messages.
```HTML

 <SemanticToastContainer position="bottom-center" maxToasts={1} />

```

## 📷 Screenshots

Before : 

https://user-images.githubusercontent.com/65409282/223626094-19a663e3-d6e8-42fe-ac5d-4756b340102b.mp4

After 👍 : 

https://user-images.githubusercontent.com/65409282/223626271-491e525f-2b11-440d-8bc6-fe6323b13b09.mp4

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -52,7 +52,12 @@
     "prefer-arrow-callback": 2,
     "prefer-const": 2,
     "prefer-promise-reject-errors": 2,
-    "prettier/prettier": 2,
+    "prettier/prettier": [
+      "error",
+      {
+        "endOfLine": "auto"
+      }
+    ],
     "sort-imports": 0,
     "sort-keys": [0],
     "sort-vars": 2,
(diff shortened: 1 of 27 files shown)


nilsnolde (approved):
LGTM

PR #107, opened 2023-03-08
Fixed [BUG] Directions Waypoint overflow #99

### LINKED ISSUE #99: [BUG] Directions Waypoint overflow
### Description

List of waypoints extend outside the box.

I think we should make the list of waypoints scrollable instead. I would like to work on it for GSoC.

### Screenshots

overflow

Ghat0tkach:
## 🛠️ Fixes Issue
Closes #99 

## 👨‍💻 Changes proposed
Since the directions were overflowing whenever we get more than 6 waypoints(checked across all devices) , so whenever there are more than 6 waypoints , the overflow automatically gets set to auto. Hence the problem is solved. 
It Also increases the height of the box to look User-Friendly.

## 📷 Screenshots

https://user-images.githubusercontent.com/59855919/223726603-70ccae72-d305-48f5-8348-6327c38519a1.mp4

### DIFF
--- a/src/Controls/Directions/Waypoints/index.jsx
+++ b/src/Controls/Directions/Waypoints/index.jsx
@@ -94,7 +94,7 @@ class Waypoints extends Component {
                 }`}
                 {...provided.droppableProps}
                 ref={provided.innerRef}
-                 style={getListStyle(snapshot.isDraggingOver, isExtendedList)}
+                style={getListStyle(snapshot.isDraggingOver, isExtendedList)}
               >
                 {waypoints.map((wp, index) => (
                   <Draggable key={wp.id} draggableId={wp.id} index={index}>
--- a/src/Controls/Directions/Waypoints/index.jsx
+++ b/src/Controls/Directions/Waypoints/index.jsx
@@ -111,14 +111,14 @@ class Waypoints extends Component {
               >
                 {waypoints.map((wp, index) => (
                   <Draggable key={wp.id} draggableId={wp.id} index={index}>
-                    {(provided, snapshot) => (
+                    {(provided_inner, snapshot_inner) => (
                       <div
-                        ref={provided.innerRef}
-                        {...provided.draggableProps}
-                        {...provided.dragHandleProps}
+                        ref={provided_inner.innerRef}
+                        {...provided_inner.draggableProps}
+                        {...provided_inner.dragHandleProps}
                         style={getItemStyle(
-                          snapshot.isDragging,
-                          provided.draggableProps.style
+                          snapshot_inner.isDragging,
+                          provided_inner.draggableProps.style
                         )}
                       >
                         <Waypoint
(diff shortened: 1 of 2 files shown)


src/Controls/Directions/Waypoints/index.jsx
```diff
@@ -34,7 +34,14 @@ const getListStyle = (isDraggingOver) => ({
   paddingBottom: 20,
   maxHeight: 350,
   height: 300,
-  oveflow: 'inherit',
+})
+const ExtendedListStyle = (isDraggingOver) => ({
```
nilsnolde, line 38: let's keep the code "style" (if there is any..) and make it camelCase

src/Controls/Directions/Waypoints/index.jsx
```diff
@@ -83,16 +90,24 @@ class Waypoints extends Component {
 
   render() {
     const { waypoints } = this.props.directions
+    const isExtendedList = waypoints.length > 6 || waypoints.length === 6
```
nilsnolde, line 93: JS surely has a concept of >=?

src/Controls/Directions/Waypoints/index.jsx
nilsnolde: in fact you could merge this with the function above right? I mean it's basically a lambda. Just accept `isExtendedList`  as an argument and make the values of `maxheight`, `height` a ternary depending on `isExtendedList`. less code and a tiny bit more readable

nilsnolde (changes requested):
Thanks, looks great! Some nits

src/Controls/Directions/Waypoints/index.jsx
Ghat0tkach: Oh yes , actually i was doing some experiments with the code .Actually  I wished to set the overflow auto only whenever the waypoints take whole up the available screen size. However this didn't go as planned . So i reverted back to the original😅

src/Controls/Directions/Waypoints/index.jsx
Ghat0tkach: Oh yes , Sorry for the inconvenience.
Will soon fix it.
Thank you

src/Controls/Directions/Waypoints/index.jsx
```diff
@@ -83,16 +81,20 @@ class Waypoints extends Component {
...
               <div
-                className={`flex flex-column`}
+                className={`flex 
+                ${
+                  isExtendedList ? 'flex-column overflow-auto' : 'flex-column'
+                }`}
                 {...provided.droppableProps}
                 ref={provided.innerRef}
-                style={getListStyle(snapshot.isDraggingOver)}
+                 style={getListStyle(snapshot.isDraggingOver)}
```
nilsnolde, line 97: you forgot to actually pass the argument;)

Ghat0tkach: but i do test the application and it worked

Ghat0tkach: iam so sorry , often in excitement . I do miss these things T-T

nilsnolde (approved):

PR #116, opened 2023-03-10
Tooltip feature added for search box in Direction and Isochrone tab

### LINKED ISSUE #115: [FEATURE] Tooltip for search box of Directions and Isochrones tab
When users search for any destination, they get a variety of choices for a destination but cannot preview the full address when they have entered the destination. We can add a tooltip for the search box in both Direction and Isochrones tabs. This tooltip will show a small preview of whatever input is given in the search box. If nothing is given input we can disable input for that time.  

It prevents from getting confusion for users because it may happen some time initial words of some addresses are almost the same and therefore it will be tedious for the user to check it by selecting it. Instead, we can show him/her a preview of input given directly.

### Screenshots

image2

### Additional information

I want to implement this feature as a part of my pre-GSOC work

MihirKp25:
## 🛠️ Fixes Issue

Closes #115

## 👨‍💻 Changes proposed

Added Popup semantic-ui-react to search boxes. Set the trigger value of Popup with the Search component. Set the content value to the userInput value given in the Search component. For no value in the search box popup will have the content "Enter Address". For some address value, it will have content as userInput

## 📷 Screenshots

image4
image5

### DIFF
--- a/src/Controls/Directions/Waypoints/Waypoint.jsx
+++ b/src/Controls/Directions/Waypoints/Waypoint.jsx
@@ -139,27 +139,35 @@ class Waypoint extends React.Component {
               </Label>
             }
           />
-          <Search
-            className={'pa2 ' + index}
-            size="small"
-            fluid
-            input={{ icon: 'search', iconPosition: 'left' }}
-            onSearchChange={this.handleSearchChange}
-            onResultSelect={this.handleResultSelect}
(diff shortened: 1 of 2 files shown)


nilsnolde (approved):
jep, thanks.

urgh, we really need this code to be more component-based. all this repetition.. not your fault of course;)

PR #117, opened 2023-03-10
Update Costing Options solved

### LINKED ISSUE #32: Update available costing options
We should update this app soon. We got a few more costing options merged recently.

hs7898753:
## 🛠️ Fixes Issue

Closes #32 

## 👨‍💻 Changes proposed
       
     Update with more costing options.

### DIFF
--- a/src/Controls/settings-options.js
+++ b/src/Controls/settings-options.js
@@ -97,6 +97,117 @@ const turnPenalityCost = {
   },
 }
 
+const maneuverPenalty = {
+  name: 'Maneuver Penalty',
+  param: 'maneuver_penalty',
+  description:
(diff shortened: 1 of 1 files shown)


hs7898753:
I added bike-share as an option in the bicycle type. 
Is it appropriate to keep it there or somewhere else?

nilsnolde:
Hm, no, this is not adding bike-share as an option. It's a slightly bigger job to add that, so let's remove anything bss-related from this PR.

I'm not happy with "profile" vs "general" settings. I didn't go through the whole thing in detail, but I saw that e.g. that only `truck` gets `gate_cost`. I'd ask you to go through the documentation and rectify what's available for which costing option. Also, even before this PR this wasn't done really correctly, but still, doesn't mean we shouldn't do a better job in the future..

PR #120, opened 2023-03-12
Remember map extent #50 [Added functionality]

### LINKED ISSUE #50: Remember map extent

harsh5902:
Closes #50 

## 🛠️ Fixes Issue
used localstorage of the user to remember the mapextent

## 👨‍💻 Changes proposed

If user moves his/her pointer on the map and moves to new location then localstorage will store his session and next time when he/she starts the app will get the same location where they left before.

You can see the center latitude and longitude on the developer tools application tab
## 📷 Screenshots

image

### DIFF
--- a/src/Map/Map.jsx
+++ b/src/Map/Map.jsx
@@ -67,9 +67,12 @@ const excludePolygonsLayer = L.featureGroup()
 
 const centerCoords = process.env.REACT_APP_CENTER_COORDS.split(',')
 let center = [parseFloat(centerCoords[0]), parseFloat(centerCoords[1])]
+let zoom_initial = 10
 
-if (localStorage.getItem('lat') && localStorage.getItem('lng')) {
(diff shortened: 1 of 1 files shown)


nilsnolde (commented):
Thanks! I'd consolidate it to a coordinate and not split up lat & lon when storing it in localstorage. Also we should capture the zoom as well as the center coordinate.

Can we name the localstorage items `last_center` & `last_zoom`?

harsh5902:
Sure I will make the necessary changes as needed

harsh5902:
@nilsnolde I have made the necessary changes you pointed out and also used localstorage for zoom level as well

image

nilsnolde:
LGTM, thanks!

PR #122, opened 2023-03-12
Increase the size of button and margin

### LINKED ISSUE #118: [BUG] less padding between buttons
On the directions tab when searching for a place, the 2 buttons (tick and cross) have very less margin between them.

Adding a little bit of padding would be nice so that on hovering over the links users don't get distracted.

### Screenshots

image

https://www.awesomescreenshot.com/video/15508755?key=87b68edcbf665c6f500b353f0a2bce98

lord2anil:
Closes #118 

## 👨‍💻 Changes proposed
image

### DIFF
--- a/src/Controls/Directions/Waypoints/Waypoint.jsx
+++ b/src/Controls/Directions/Waypoints/Waypoint.jsx
@@ -169,33 +169,39 @@ class Waypoint extends React.Component {
               />
             }
           />
-          <Popup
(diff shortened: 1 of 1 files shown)


nilsnolde:
Looks good, thanks! WDYT @Ananya2001-an ?

Ananya2001-an (approved):
LGTM👍

PR #125, opened 2023-03-13
Closes #123 

### LINKED ISSUE #123: [BUG] Search Results Trapped Inside Scrolling
I noticed that the search results still become part of the scrolling after it extends in size at the 6+ waypoint.

Capture

I am suggesting to instead set the min height to the 6th waypoint and increase the height per waypoint. Also to move the scrolling to the outer div so that the search results will not become part of the scrolling inside the list of waypoints.

Capture2

### Screenshots

_No response_

rakeshhotker:
## 🛠️ Fixes Issue 
 Closes #123 

## 👨‍💻 Changes proposed
In the semantic-ui-css package, added height: 200px and overflow-y:scroll for the .ui.search > .results selector.
Since it's a change in the node-modules, to add the patch followed the following steps:
1) Install patch-package: "npm install patch-package -D"
2) add postinstall script:  "scripts": {
  "postinstall": "patch-package"
 } , added the above code in package.json file
 3) Run patch-package to create a .patch file:
     "npx patch-package semantic-ui-css" 
 4) Apply your fix via npm package manager: "npm install"

## 📄 Note to reviewers
Please suggest any changes required

## 📷 Screenshots
This is how it looks.
Screenshot from 2023-03-13 23-13-41

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "throttle-debounce": "^5.0.0"
   },
   "scripts": {
+    "postinstall": "patch-package",
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
@@ -87,6 +88,7 @@
     "gh-pages": "^4.0.0",
     "husky": "^8.0.2",
     "lint-staged": "^13.0.3",
+    "patch-package": "^6.5.1",
     "prettier": "^2.7.1",
     "redux-devtools-extension": "^2.13.8"
   }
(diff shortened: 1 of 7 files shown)


nilsnolde (commented):
Hmm, patching a dependency is a last resort sort of thing. This WILL break, not tmrw, but very surely in the next year. I'd rather not have a "time bomb" in the repo.

Is there no way to add those attributes to the element(s) needing it?

rakeshhotker:
i couldn't find one. But there's another PR from @tommach591  which claims to fix it. So am closing this one. Thanks. @nilsnolde.

rakeshhotker:
I will close it if you think @tommach591 PR is good enough to solve the issue @nilsnolde.

PR #126, opened 2023-03-14
Added title property to heightgraph-toggle element to show "Distance …

nomandhoni-cs:
## 📄 Note to reviewers
- Added title property to heightgraph-toggle element to show "Distance Graph" tooltip in the icon in bottom right.
- It was little bit confusing what this button does, with tooltip it will improve User Experience.

### After
Screenshot_52

### DIFF
--- a/src/Map/Map.jsx
+++ b/src/Map/Map.jsx
@@ -307,6 +307,8 @@ class Map extends React.Component {
     })
     this.hg.addTo(this.map)
     const hg = this.hg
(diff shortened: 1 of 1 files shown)


nomandhoni-cs:
## After
https://user-images.githubusercontent.com/92979541/225015247-0c657a70-f553-4475-8823-1efc6030fd2a.mp4

nilsnolde:
I didn't test it, but I trust it works as you say:)

nomandhoni-cs:
Thanks for merging!

Issue #121, opened 2023-03-12
[BUG] Debounce Isochrone API calls when adjusting interval settings by dragging the sliders


wcedmisten:
### What is the problem you are facing
Currently when these sliders get adjusted by dragging them (not by clicking on a value), the frontend makes a request for each interval between the previous value and the new value. This results in the page lagging as the ~60 requests return sequentially and each redraw the polygon. This also puts extra strain on the server from this many requests being made in quick succession.

### What is your suggested solution
Adjusting the slider should not send a request with the new settings until the mouse click is released.

### Screenshots

![image](https://user-images.githubusercontent.com/4565200/224553489-50074f71-e7d8-4f51-a0d9-973bfe7b1849.png)

![image](https://user-images.githubusercontent.com/4565200/224553634-75e689c8-33f5-4173-8bb9-9fc052d74e00.png)

nilsnolde:
Oh yes, good one!

sandstone991:
Hello @nilsnolde you might remember me from my previous unannounced PRs, so I'm going to ask to be assigned first this time, can I work on this?

nilsnolde:
Sure, thanks!

PR #135, opened 2023-03-17
Closes (#121)

sandstone991:
## 🛠️ Fixes Issue
Debounce Isochrone API calls  #121 

## 👨‍💻 Changes proposed

Changed the UNSAFE_ComponentWillUpdate lifecycle which was ran on every change to see if the sliders values had changed to send a new request, made the component call a debounced fetch function on change instead.
As for the sliders they make request only on the mouseUp event (user let go of the slider).

## 📄 Note to reviewers

I encounted two bugs (unrelated to the task) while working on this component.

- The max value of the range is set to 120 
image
which gives error code 151 when calling Valhalla's API (Exceeded max time 100) so I'm guessing the max value has to change to 100 here

- Some requests return 504 which shows the user this messege 
image
which is unhelpful IMO

I'm also dissatisfied with how async calls are done within this component, given unfortuante enough cirumstances a race condition like this can happen 
image

but I think that have to wait until the components are rewritten into functional components

Also please consider this a pre-gsoc pr

### DIFF
--- a/src/Controls/Isochrones/Waypoints/index.jsx
+++ b/src/Controls/Isochrones/Waypoints/index.jsx
@@ -1,7 +1,6 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
-// import * as R from 'ramda'
 
 import { Search, Form, Popup, Icon, Label, Accordion } from 'semantic-ui-react'
 import { Slider } from '@mui/material'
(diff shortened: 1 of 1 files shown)


sandstone991:
@nilsnolde kindly review this

src/Controls/Isochrones/Waypoints/index.jsx
```diff
@@ -233,7 +237,7 @@ class Waypoints extends Component {
                 loading={isFetching}
                 results={geocodeResults}
                 value={userInput}
-                onKeyPress={(event: React.KeyboardEvent) => {
```
chrstnbwnkl, line 236: thanks for catching this, shouldn't have made it into master in the first place

src/Controls/Isochrones/Waypoints/index.jsx
```diff
@@ -85,18 +88,18 @@ class Waypoints extends Component {
...
-    const nextMaxRange = nextProps.isochrones.maxRange
-    const nextInterval = nextProps.isochrones.interval
-    if (
-      !R.equals(maxRange, nextMaxRange) ||
-      !R.equals(interval, nextInterval)
-    ) {
-      dispatch(makeIsochronesRequest())
-    }
-  }
+  // componentDidUpdate(nextProps) {
```
chrstnbwnkl, line 91: don't just uncomment unused code, remove it instead

src/Controls/Isochrones/Waypoints/index.jsx
```diff
@@ -1,7 +1,7 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
-import * as R from 'ramda'
+// import * as R from 'ramda'
```
chrstnbwnkl, line 4: should be removed as well

chrstnbwnkl (changes requested):
Just some minor comments

src/Controls/Isochrones/Waypoints/index.jsx
sandstone991: sure, no problem

PR #142, opened 2023-03-21
if else statements to switch

### LINKED ISSUE #141: Modify the customSlider handle change function
### Screenshots

_No response_

### Additional information

 const handleChange = (value) => {
      //reset
      switch (true) {
        case isNaN(value):
          value =
            profile === 'truck'
              ? settingsInitTruckOverride[option.param]
              : settingsInit[option.param];
          break;
        case value < min:
          value = min;
          break;
        case value > max:
         value = max;
          break;
        default:
        break
      }
    
      setSliderVal(parseFloat(value));
    
      debounce(
        300,
        handleUpdateSettings({
          name: option.param,
          value: parseFloat(value),
        })
      );
    };

DevMarioNan:
turning the customSlider handle change function from an if else if statement to a switch statement

#141# 🛠️ Fixes Issue

 Closes #141 

## 👨‍💻 Changes proposed

in the CustomSlider.jsx  component the handleChange function have some if else statements which slows down the process by a little bit

so i've changed it to a switch statement

the switch statement is faster in most cases when compared to if-else, but significantly faster only when the number of conditions is large. And It would help to make it more readable

## 📄 Note to reviewers

the switch statement is faster in most cases when compared to if-else, but significantly faster only when the number of conditions is large. And It would help to make it more readable

### DIFF
--- a/src/components/CustomSlider.jsx
+++ b/src/components/CustomSlider.jsx
@@ -20,16 +20,22 @@ const CustomSlider = (props) => {
 
   const handleChange = (value) => {
     // reset
-    if (isNaN(value)) {
-      value =
+
+    switch(true){
+      case isNaN(value):
+        value =
         profile === 'truck'
           ? settingsInitTruckOverride[option.param]
           : settingsInit[option.param]
-    }
-    if (value < min) {
-      value = min
-    } else if (value > max) {
-      value = max
+        break;
+      case value < min:
+        value = min
+        break;
+      case value > max:
+        value = max
+        break;
+      default:
+        break;
     }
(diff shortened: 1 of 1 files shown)


Ananya2001-an (commented):
Hey @DevMarioNan thanks for trying to contribute 👍🏼  but I feel there's no need to create a switch case because it unnecessarily adds more lines of code. If there were many conditions then we might prefer it but just for 3 conditions it doesn't make any sense tbh.

nilsnolde:
Yeah agreed. There’s also no actual variable to switch over. It does need a “else if” but not necessarily a switch.

Issue #67, opened 2023-02-23
GSoC 2023 - README!


nilsnolde:
Let's try this again:

**READ AND FOLLOW THIS CAREFULLY IF YOU WANT TO BE CONSIDERED**

We highly appreciate the volume of people who want to be considered for our project idea for GSoC 2023. However, please **hold off raising PRs**. There is an application process behind GSoC and we will follow it. You will be notified when the application process for contributors is opened (either by us and/or Google) and **only then** should you apply. 

We will close any PRs which will be opened after this issue and will not consider the person raising it. Consider it a first test.

nilsnolde:
I wrote down a few things which should be followed this year for GSoC:

https://github.com/gis-ops/valhalla-app/wiki/GSoC-2023

Good luck!

nilsnolde:
For anyone having open PRs: I'll get to them soon, likely on the weekend. Already thanks for taking care of so many issues! 

Best if everyone subscribes to this issue so you updates when some more info becomes available.

AFZL210:
I have a question regarding discussions. Is there any public channel for contributors?

nilsnolde:
So, for the proposal, I'd suggest to focus on:
- testing end-to-end including a Github Actions workflow which runs on PRs, so that the most important functionality is preserved. This might need a bit of more setup, e.g. getting a local Valhalla setup on Github Actions. Explicitly, we **don't want unit tests**, it's an unnecessary burden for the project (remember, after GSoC is gone, it'll still mostly be me who's maintaining all this..)
- re-write using functional components instead of class-based as now. Don't worry about the exact design/architecture of it for now, it's part of the project 
- being able to link/integrate better to openstreetmap.org, including, but not limited to (own suggestions welcome):
  - a tool button to jump to the same map extent as currently active
  - re-create a route on openstreetmap.org for the active profile (note, not all profiles are supported for Valhalla on osm.org)
  - include an entry in the left-click menu to call Valhalla's `/locate` endpoint to retrieve the OSM ID of the way being clicked on so we can see the way's properties on osm.org (first `/locate` result is enough, as there might be many edges being returned)

AdamJamiu:
HI @nilsnolde I would love to work **_re-create a route on openstreetmap.org for the active profile (note, not all profiles are supported for Valhalla on osm.org)_**. please how can i get started 😊

nilsnolde:
Sorry, gotta be strict here: that will be part of the GSoC project.

[...]

PR #149, opened 2023-03-24
Height-graph width reduced

### LINKED ISSUE #136: [BUG] HeightGraph Y-axis goes behind Directions Tab
Users when trying to find Directions using the Directions tab and at the same time want to look at the height graph, then some part of the graph is not visible. Also when the user closes that tab, the graph tab also closes automatically. 

We can improve user experience by solving this problem. We can either reduce the size of the graph permanently or we can reduce its width only for the time when the tab is open. This thing can be achieved by managing states. I think the latter idea would be better, though managing states is a bit difficult for me here. If anyone has already managed states in it please tell me. 

### Screenshots

image8

MihirKp25:
## 🛠️ Fixes Issue
Closes #136 

## 👨‍💻 Changes proposed

I have reduced the width of height graph from 0.9 to 0.75 in Map.jsx file. I tried to change this width dynamically but it didn't end well. But still it looks pretty nice to me and now we can easily see y axis of height graph.

## 📄 Note to reviewers

If you find a way of reducing the width dynamically whenever the direction tab closes please tell me :)

## 📷 Screenshots
image9

### DIFF
--- a/src/Controls/Directions/Waypoints/Waypoint.jsx
+++ b/src/Controls/Directions/Waypoints/Waypoint.jsx
@@ -139,6 +139,7 @@ class Waypoint extends React.Component {
               </Label>
             }
           />
+
           <Popup
             content={userInput.length === 0 ? 'Enter Address' : userInput}
             size="tiny"
--- a/src/Controls/Directions/Waypoints/Waypoint.jsx
+++ b/src/Controls/Directions/Waypoints/Waypoint.jsx
@@ -139,8 +139,6 @@ class Waypoint extends React.Component {
               </Label>
             }
           />
-
-
           <Popup
             content={userInput.length === 0 ? 'Enter Address' : userInput}
             size="tiny"
--- a/src/Controls/Directions/Waypoints/Waypoint.jsx
+++ b/src/Controls/Directions/Waypoints/Waypoint.jsx
@@ -139,27 +139,35 @@ class Waypoint extends React.Component {
               </Label>
             }
           />
-          <Search
-            className={'pa2 ' + index}
-            size="small"
-            fluid
-            input={{ icon: 'search', iconPosition: 'left' }}
-            onSearchChange={this.handleSearchChange}
-            onResultSelect={this.handleResultSelect}
-            resultRenderer={this.resultRenderer}
-            type="text"
-            showNoResults={false}
-            open={this.state.open}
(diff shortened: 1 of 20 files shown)


chrstnbwnkl:
@MihirKp25 could you please resolve existing conflicts so we can have a look at your PR?

MihirKp25:
I have resolved conflicts. Sorry for inconvenience.

nilsnolde (changes requested):
The whitespace-only changes seem irrelevant, I’d actually expect prettier to complain.. Can you revert those?

src/Controls/Directions/Waypoints/Waypoint.jsx
```diff
@@ -139,7 +139,6 @@ class Waypoint extends React.Component {
               </Label>
             }
           />
-
```
chrstnbwnkl, line 142: please check your PR again for unnecessary, whitespace-only changes.

chrstnbwnkl (changes requested):

src/Controls/Directions/Waypoints/Waypoint.jsx
MihirKp25: Sorry I misinterpreted the last change request and added spaces in this file. I removed those changes. Please let me know if there are any other unnecessary changes I have made. And sorry for the inconvenience.

nilsnolde (approved):
Fine for me, thanks. I'll let @chrstnbwnkl merge if he's ok with this.

MihirKp25:
I will try to make your suggested changes asap.

chrstnbwnkl:
@MihirKp25 can you push your latest changes? The functionality in the video looks pretty close to what we want to achieve, right?

MihirKp25:
Changes are pushed @chrstnbwnkl

nilsnolde:
This looks pretty good, thanks @MihirKp25 ! 

You can find more info on the proposal here: https://wiki.openstreetmap.org/wiki/Google_Summer_of_Code/2023 and here: https://github.com/gis-ops/valhalla-app/issues/67. If you follow those guidelines you should be good. We have (un)fortunately too many applicants to do extensive proposal review. I suggest you upload it to GSoC, we'll surely read and consider it.

nilsnolde (approved):

PR #150, opened 2023-03-24
resolved issue Include OSM link somewhere #54

### LINKED ISSUE #54: Include OSM link somewhere
For debugging it'd be super helpful if we could quickly switch to openstreetmap.org. 

For now, it'd be enough (for me) to have a tool button in the right vertical toolbar to just switch to (more or less) the same map extent on osm.org. We get a lot of issues with links of our web app where we'd like to see the tagging quickly.

Eventually we could add a button in the Directions dialog to generate the same route on osm.org, but less of a priority.

Also, it'd be cool to optionally issue a locate request in the left-click menu to jump to the first way's OSM link.

harsh5902:
## 🛠️ Fixes Issue
issue resolved #54 

## 👨‍💻 Changes proposed
added button "open OSM" to valhalla web app to link OSM with same map extent and zoom level.

## 📷 Screenshots
image

### DIFF
--- a/src/Map/Map.css
+++ b/src/Map/Map.css
@@ -1,20 +1,9 @@
-.osm-button {
-    position: absolute;
-    bottom: 20px;
-    right: 60px;
-    z-index: 1000;
-    background-color: #2185D0;
-    color: #FFFFFF;
-    border: 1px none #333;
-    border-radius: 4px;
-    padding: 8px 16px;
-    font-size: 14px;
-    font-weight: bold;
-    text-align: center;
-    cursor: pointer;
-    box-shadow: 0 1px 3px rgba(0,0,0,0.12), 0 1px 2px rgba(0,0,0,0.24);
-    transition: all 0.3s cubic-bezier(.25,.8,.25,1);
-  }
+#osm-button {
+  position: absolute;
+  bottom: 20px;
+  right: 60px;
+  z-index: 1000;
+}
 
 /* defining the container styles the map sits in */
 .map-style{
--- a/src/Map/Map.jsx
+++ b/src/Map/Map.jsx
@@ -1033,7 +1033,11 @@ class Map extends React.Component {
       <React.Fragment>
         <div>
           <div id="map" className="map-style" />
-          <button className="osm-button" onClick={this.handleOpenOSM}>
+          <button
+            className="ui primary button"
+            id="osm-button"
+            onClick={this.handleOpenOSM}
+          >
             Open OSM
           </button>
         </div>


package.json
```diff
@@ -14,6 +14,7 @@
     "jquery-ui-bundle": "^1.12.1-migrate",
     "json-format": "^1.0.1",
     "leaflet": "^1.9.3",
+    "leaflet-easybutton": "^2.4.0",
```
nilsnolde, line 17: seems like this isn't used right? which is really good, since it's not maintained since 4 years, i.e. we wouldn't want it, another time bomb:)

src/Map/Map.jsx
```diff
@@ -42,6 +42,24 @@ const style = {
   height: '100vh',
 }
 
+const button_style = {
```
nilsnolde, line 45: it's ok for now to define in JS but eventually we should move all that stuff to css files

harsh5902:
sure I will make changes..

package.json
harsh5902: No, actually this bug was left by me, my mistake😢
I first tried to approach this problem using leaflet easybutton.

nilsnolde:
Thanks, works for me. WDYT @chrstnbwnkl?

harsh5902:
@nilsnolde I have made the necessary changes as you suggested

And a question GSOC related for how much week should I keep deadline of project on my proposal right now I have kept it for 10 weeks.

src/Map/Map.jsx
chrstnbwnkl: I think it could be part of the project to determine how styles are managed! CSS files on a per-component basis are surely an option but there's plenty more. But yeah, consistency is key :-)

nilsnolde: Maybe I should just stay out of the review for web dev stuff😄

harsh5902: yes, as more css could be the part I created new css file for Map component @chrstnbwnkl

src/Map/Map.css
```diff
@@ -0,0 +1,23 @@
+.osm-button {
```
chrstnbwnkl, line 1: Adding that class here with all those custom styles is not necessary, please take a look at the UI component library we're using. Hint: take a look at the Directions/Isochrone button in the top right when the sidebar is collapsed.

chrstnbwnkl (changes requested):

harsh5902:
image
Have a look now @chrstnbwnkl

src/Map/Map.css
```diff
@@ -0,0 +1,12 @@
+#osm-button {
```
chrstnbwnkl, line 1: This can go now.

chrstnbwnkl (changes requested):
Please remember to remove any unused code.

PR #152, opened 2023-03-24
Fix route shown even if input is altered

### LINKED ISSUE #144: Route shown even if waypoint input is altered
Right now when we change the input of any waypoint, the route is still shown.

Any change made to input should remove the route automatically just as what happens when we click the cross icon to remove the waypoint manually.

### Screenshots

https://user-images.githubusercontent.com/55504616/227126913-7ffca138-cc71-45cc-a84b-65fffc6f6dd3.MOV

Sheikh-JamirAlam:
## 🛠️ Fixes Issue
Fixes #144 

## 👨‍💻 Changes proposed
Now we can alter the waypoint input field and the route will disappear.

## 📄 Note to reviewers
Please see the attached video.

## 📷 Screenshots

https://user-images.githubusercontent.com/48979376/227571796-9e8fa0de-408b-48f4-95f9-5c68375e7abe.mp4

### DIFF
--- a/src/actions/directionsActions.js
+++ b/src/actions/directionsActions.js
@@ -371,7 +371,6 @@ export const doAddWaypoint = (doInsert) => (dispatch, getState) => {
     isFetching: false,
     userInput: '',
   }
-  //console.log(emptyWp)
   if (doInsert) {
     dispatch(insertWaypoint(emptyWp))
   } else {
--- a/src/actions/directionsActions.js
+++ b/src/actions/directionsActions.js
@@ -311,6 +311,13 @@ export const doRemoveWaypoint = (index) => (dispatch, getState) => {
   dispatch(updatePermalink())
 }
 
+export const isWaypoint = (index) => (dispatch, getState) => {
+  const waypoints = getState().directions.waypoints
+  if (waypoints[index].geocodeResults.length > 0) {
+    dispatch(clearRoutes(VALHALLA_OSM_URL))
+  }
+}
+
 export const highlightManeuver = (fromTo) => (dispatch, getState) => {
   const highlightSegment = getState().directions.highlightSegment
   // this is dehighlighting
@@ -364,6 +371,7 @@ export const doAddWaypoint = (doInsert) => (dispatch, getState) => {
     isFetching: false,
     userInput: '',
   }
+  //console.log(emptyWp)
   if (doInsert) {
     dispatch(insertWaypoint(emptyWp))
   } else {
--- a/src/Controls/Directions/Waypoints/Waypoint.jsx
+++ b/src/Controls/Directions/Waypoints/Waypoint.jsx
@@ -7,10 +7,8 @@ import {
   updateTextInput,
   fetchGeocode,
   makeRequest,
-  clearRoutes,
+  isWaypoint,
 } from 'actions/directionsActions'
-
-import { VALHALLA_OSM_URL } from 'utils/valhalla'
 
 import { zoomTo } from 'actions/commonActions'
 import { isValidCoordinates } from 'utils/geom'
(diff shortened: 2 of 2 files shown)


Ananya2001-an (commented):
Hey @Sheikh-JamirAlam Thanks for the quick PR. But actually this is not going to work, every time we make changes to the input it will clear the routes which means nothing is getting shown. Did you test it properly? What we want is that when someone starts altering a waypoint's input value then that marker as well as the route associated to that point should get removed. I hope you get it now. Feel free to ask if you still have any doubt :)

Sheikh-JamirAlam:
@Ananya2001-an Sorry I think I don't get it. What do you mean by altering a waypoint's input value?

Ananya2001-an:
> @Ananya2001-an Sorry I think I don't get it. What do you mean by altering a waypoint's input value?

In the screenshot in issue #144 I showed that once we start changing the input value of a waypoint the marker disappears but the route is still there. So what we want is to remove it if after drawing a route someone starts changing the input (waypoint). But your code is not letting routes getting drawn on the map at all. It will just clear up every time someone types into the search box. Hope u get it now. Please play with the code a bit to understand which code is affecting which part of the app.

Sheikh-JamirAlam:
@Ananya2001-an Sorry for the late response, I think I have managed to fix it. Am sharing the video and commint my changed now. Also I created a function and named it `isWaypoint`, hope the naming is fine.

https://user-images.githubusercontent.com/48979376/227718462-08836a77-84ab-4e46-b1ac-0df51fc369bd.mp4

src/actions/directionsActions.js
```diff
@@ -311,6 +311,13 @@ export const doRemoveWaypoint = (index) => (dispatch, getState) => {
   dispatch(updatePermalink())
 }
 
+export const isWaypoint = (index) => (dispatch, getState) => {
+  const waypoints = getState().directions.waypoints
+  if (waypoints[index].geocodeResults.length > 0) {
+    dispatch(clearRoutes(VALHALLA_OSM_URL))
```
Ananya2001-an, line 317: This will clear all the routes, we just want to remove the one that was associated with that marker.

Ananya2001-an:
> @Ananya2001-an Sorry for the late response, I think I have managed to fix it. Am sharing the video and commint my changed now. Also I created a function and named it `isWaypoint`, hope the naming is fine.

Don't stress about it. It's not a simple issue. Try to work according to your pace. You are already doing good 👍🏼

src/actions/directionsActions.js
nilsnolde: I think that's fine no? Are you thinking to just remove a "section" of the route? So if there's 3 waypoints and a user alters the very last one, you'd like to just remove the route's section between the 2nd and last waypoint? 

That'd be possible but IMHO unnecessary as we'd send a new route request with _all_ waypoints anyways (just to keep things simple).

Ananya2001-an: Yeah, I guess you are right. Maybe we should keep it simple. I think this is good enough. I was making it complex in terms of UX but I guess it's not important unless it turns out to be a concern in the future.

Sheikh-JamirAlam: @nilsnolde @Ananya2001-an So, will this get merged?

src/actions/directionsActions.js
```diff
@@ -364,6 +371,7 @@ export const doAddWaypoint = (doInsert) => (dispatch, getState) => {
     isFetching: false,
     userInput: '',
   }
+  //console.log(emptyWp)
```
nilsnolde, line 374: Pls remove comments

nilsnolde (commented):
Got my approval, just remove the comments. Thanks again!

Ananya2001-an (approved):
Looks good! 👍🏼 Just remove the comment as Nils mentioned above.

src/actions/directionsActions.js
Sheikh-JamirAlam: Done

PR #153, opened 2023-03-25
fixes Inconsistent height graph coloring #57

### LINKED ISSUE #57: Inconsistent height graph coloring
https://valhalla.openstreetmap.de/directions?profile=bicycle&wps=13.376197814941408%2C52.547776790183214%2C13.38589668273926%2C52.55713840780446%2C13.393707275390627%2C52.5650153081717%2C13.390789031982424%2C52.56655304975672%2C13.389244079589846%2C52.57029962885746

In many cases, elevation change and steepness are contradicting, see this example:
grafik

Binabh:
## 🛠️ Fixes Issue

Closes #57 

## 👨‍💻 Changes proposed

- new approach in building the geojson so that the inconsistency in coloring in height graph is removed

## 📄 Note to reviewers

Please let me know if any changes/improvements required

## 📷 Screenshots

checked on [this](https://github.com/gis-ops/valhalla-app/issues/57#issue-1573894591) case

image

checked on [this](https://github.com/gis-ops/valhalla-app/issues/57#issuecomment-1428556845) case
image

### DIFF
--- a/src/utils/heightgraph.js
+++ b/src/utils/heightgraph.js
@@ -65,13 +65,22 @@ export const buildHeightgraphData = (coordinates, rangeHeightData) => {
         declineTotal += rise * -1
       }
 
+      // IDEA FOR BUILDING GEOJSON FOR HEIGHTGRAPH:
+      // Basic idea is keepon adding points to current LineString as long as they have same heightClass
+      // If we sense that the heightClass has changed, we add current LineString as one feature in GeoJSON
+      // new heightClass new LineString
+      // height graph works by taking a geojson with collection of LineStrings and coloring them based on heightClass property
+
+      // this checks if heightClass has changed and we need to add current LineString as one feature in GeoJSON
       if (previousHeightClass !== heightClass) {
+        // since at this point we have a change in height class this will be the last point in current LineString
         LineStringCoordinates.push([
           coordinates[index][0],
           coordinates[index][1],
           rangeHeightData[index][1],
         ])
 
+        // add current LineString as one feature in GeoJSON
         features.push({
           type: 'Feature',
           geometry: {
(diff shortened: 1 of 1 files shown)


nilsnolde (commented):
Ah nice! That's a very long-standing issue and been bothering quite a bit (though apparently not to the point to attempt a fix..). Thanks a lot @Binabh, looks good! 

Just one further ask: could you leave some comments in the code what this does? I know there were no comments before either, but that's also bothering. The logic is not apparent to me tbh, e.g. what `attributeType` does and why we're pushing the same coordinates twice if the heighClass changed. I'm also lacking the context without looking into the code and see what the heigh graph package needs and does. So for reviewing it'd be much more convenient to read a comment what it does to give the context. Hope that makes sense.

Binabh:
sure @nilsnolde I will try to explain the best I can and best I have understood. I will be adding commit with comments in few hours.
Thank You for your suggestion and review !

Binabh:
@nilsnolde I have tried to add some explanation. I have to admit that documentation is not my strong part. I further explain any particular line if required.

Also, one major difference between previous and current code is that previously we were making new linestrings if current heightClass is different than previous but what we actually needed to do is end current linestring and start a new one as soon as we see change in height class. Hope this makes sense.

nilsnolde (approved):
Right, it's not super well written tbh, but it's really ok, I understand it now better for sure, thanks. I'd generally try to work a bit more on docs skills, it's really important for devs IMO, it makes for much more seamless collaboration with others. And remember to annotate/comment your PRs for any non-trivial things to put context for reviewers, I'm always very happy to see that so it's less time for me to review:)

Anyways, I'm really grateful you solved this! And IMO pretty elegantly, pretty much without any additional lines, I like that very much:)

Binabh:
Sure, @nilsnolde thank you for your suggestion.
Also thank you for the review/merge and this awesome project.

Issue #138, opened 2023-03-19
[FEATURE] Use User Feedback and Integrate weather forecast feature


Akshat171:
### What kind of feature you are suggesting?
we can add a Feedback form and integrate a weather forecast feature for all places. 

### Does it solve any problem?
Collecting user feedback is a great way to learn what is working well and what needs to be improved in your map application. Consider implementing a feedback system or survey to gather user feedback and make improvements based on that feedback Consider integrating a map application with another service that users may find useful, such as weather forecasts. This can add value to your map application and keep users engaged.

hs7898753:
Weather conditions can change quickly and may not have a significant impact on the route or the user's travel plans.
but yes IMO it may be better to provide a weather forecast as an optional feature that users can access within the app, rather than integrating it into routing.

Akshat171:
It can affect users' travel plans it provides valuable information to users about current and future weather conditions in a particular location. This information can be critical for making decisions related to travel, outdoor activities, and safety.

For example, if a user is planning a road trip, they can use the weather forecast to determine the best route based on current and predicted weather conditions. If there is a chance of severe weather, such as thunderstorms or heavy snowfall, users can adjust their travel plans accordingly.

hs7898753:
I agree with your point. But, integrating weather forecasts into routing may not always be necessary. Ex. if a user is only traveling a short distance, weather information may not impact their route or travel plans significantly. Moreover, integrating weather forecasts into routing may increase the complexity of the app, making it harder for users to use and navigate.

Therefore, providing a weather forecasts as an optional feature that users can access within the app is a better approach. 
WDYT @nilsnolde

nilsnolde:
Even though you’re right, integrating weather forecast is nice and even nicer to actually include it in the routing algo as costing instead of pure informational as a visual layer, this is a bit over the top for now. This is still mostly a demo app for a routing engine we maintain. In case it should ever go further than that, we can consider this, similar to #133.

Akshat171:
@hs7898753 this application is not for the user who going to travel short distances. we have to think from both point of view so I think it is the common or main feature of every map application.

Akshat171:
@nilsnolde Please review this feature, is there anything I can improve on?

Pratham271:
We can also suggest the alternate route if the weather is too bad at the primary route as we are already adding the weather forecast condition

PR #155, opened 2023-03-25
weather forcast feature

Akshat171:
## 🛠️ Fixes Issue
 integrating a weather forecast feature.

https://github.com/gis-ops/valhalla-app/issues/138#issue-1630809203

## 👨‍💻 Changes proposed
Now we can see the weather forecast and can add some value to this application 

## 📄 Note to reviewers

PFA (video) below

## 📷 Screenshots
https://user-images.githubusercontent.com/81281246/227740008-e9c5512f-954b-44e8-a379-c45d4c6cd7ce.mp4


nilsnolde:
Maybe I should've been more explicit in https://github.com/gis-ops/valhalla-app/issues/138#issuecomment-1475212635. IMO it's not very useful. I think it's easy enough to open another browser window or look on the phone's weather app. Also, this is a third-party service which requires an API key (probably commercial service with a freemium API key right?). You hard-coded that into the source code of a OSS repo, everyone can see and use that now, I hope you didn't attach your credit card;) 

All in all, sorry, but it's not for us.

PR #156, opened 2023-03-26
Button created to locate ourselves: issue resolved #131

### LINKED ISSUE #131: [FEATURE] Adding current location button
I would like to add a button which will move the map to user's location. 
Usually user tends to locate himself/herself or try to know the distance from his location to other.

### Screenshots

_No response_

### Additional information

_No response_

harsh5902:
## 🛠️ Fixes Issue
issue #131 resolved

## 👨‍💻 Changes proposed
Created a button to move map to our location
## 📷 Screenshots
image

### DIFF
--- a/src/Map/Map.jsx
+++ b/src/Map/Map.jsx
@@ -132,6 +132,8 @@ class Map extends React.Component {
       isHeightLoading: false,
       locate: [],
     }
+    this.map = null
+    this.handleLocateMe = this.handleLocateMe.bind(this)
   }
 
   // and once the component has mounted we add everything to it
@@ -831,6 +833,13 @@ class Map extends React.Component {
     window.open(osmURL, '_blank')
   }
 
+  handleLocateMe = () => {
+    navigator.geolocation.getCurrentPosition((position) => {
+      const { longitude, latitude } = position.coords
+      this.map.setView([latitude, longitude], 12)
+    })
+  }
+
   render() {
     const { activeTab } = this.props
     const MapPopup = (isInfo) => {
@@ -1040,6 +1049,14 @@ class Map extends React.Component {
             onClick={this.handleOpenOSM}
           >
             Open OSM
+          </button>
+          <button
+            className="ui primary button"
+            id="locateme"
+            onClick={this.handleLocateMe}
+            title="Move to your location"
+          >
+            <Icon name="map marker alternate" />
           </button>
         </div>
         <div>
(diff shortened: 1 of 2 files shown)


src/Map/Map.jsx
```diff
@@ -831,6 +833,13 @@ class Map extends React.Component {
     window.open(osmURL, '_blank')
   }
 
+  handleLocateMe = () => {
+    navigator.geolocation.getCurrentPosition((position) => {
+      const { longitude, latitude } = position.coords
+      this.map.setView([latitude, longitude], 12)
```
nilsnolde, line 839: this really needs some comments. always try to put yourself in the shoes of reviewers: we seldomly touch this codebase and just see a diff and I want to avoid opening an IDE for review, so I have zero context. 

where's `navigator` coming from? `getCurrentPosition` takes the lambda, passes `position` to it and it sets the current map view right?

src/Map/Map.jsx
```diff
@@ -132,6 +132,8 @@ class Map extends React.Component {
       isHeightLoading: false,
       locate: [],
     }
+    this.map = null
```
nilsnolde, line 135: why this?

nilsnolde (commented):
Can't really judge it that well currently, as there's no comments explaining the code changes. It's good practice to give reviewers some context so reviewing doesn't require opening IDE and/or researching some API.. 

Other than that, I'd be fine with the concept I think, though not super happy with the button and its position. Is it possible to make it a leaflet control/button and integrate it into the right "sidebar" of buttons? e.g. above/below the zoom buttons.

So what happens now, is that when someone clicks on that button, the browser will ask for location permission right? What exactly happens if permission is denied?

harsh5902:
> Can't really judge it that well currently, as there's no comments explaining the code changes. It's good practice to give reviewers some context so reviewing doesn't require opening IDE and/or researching some API..
> 
> Other than that, I'd be fine with the concept I think, though not super happy with the button and its position. Is it possible to make it a leaflet control/button and integrate it into the right "sidebar" of buttons? e.g. above/below the zoom buttons. Tbh, I'm not too familiar with leaflet, but I'd imagine it's not hard?
> 
Ok I will make the necessary changes
> So what happens now, is that when someone clicks on that button, the browser will ask for location permission right? What exactly happens if permission is denied?
Yes, it will ask for user consent first whether he/she is okay with sharing the location or not and then move to the map to the user's location.

nilsnolde:
And 

> What exactly happens if permission is denied?

Simply nothing happens then right?

harsh5902:
> And
> 
> > What exactly happens if permission is denied?
> 
> Simply nothing happens then right?

Yeah!! nothing will happen.

PR #160, opened 2023-03-26
Adjusting map bounds (removing gray area)

hs7898753:
## 🛠️ Fixes Issue

Currently, map can be dragged beyond its limit. showing an extra gray area.

## 📷 Screenshots

image


nilsnolde (approved):
Jep, makes sense IMO. Still, a few comments to explain the change in behavior would be good generally.

PR #161, opened 2023-03-28
Costing Option Display

### LINKED ISSUE #147: [BUG] Simplify Costing Options Display

While working on updating costing options. 
I noticed the costing options display should be clear and easy to navigate, even when there are many options available. 

To simplify the display, I suggest showing only the general settings by default, and moving the extra settings into a dropdown menu. This will allow users to easily access the settings they need, without being overwhelmed by too much information at once.

Or what you suggest? 

### Screenshots

_No response_

hs7898753:
## 🛠️ Fixes Issue

Closes #147 

## 👨‍💻 Changes proposed
 
I suggest showing only the general settings by default and moving the extra settings into a dropdown menu. This will allow users to easily access the settings they need, without being overwhelmed by too much information at once.

## 📷 Screenshots

https://user-images.githubusercontent.com/112114562/228326988-10db5986-75ac-4d8d-b796-b180192ff34b.mp4

### DIFF
--- a/src/Controls/settings-panel-css.css
+++ b/src/Controls/settings-panel-css.css
@@ -1,0 +1,44 @@
+/* Custom styles for the panel */
+.panel-container {
+  display: flex;
+  flex-direction: column;
+}
+
+.panel-section {
+  margin-bottom: 1rem;
+}
+
+.general-settings {
+  order: 1;
+}
+
+.extra-settings {
+  order: 2;
+}
+
+.panel-header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  margin-bottom: 0.5rem;
+}
+
+.panel-header h4 {
+  margin: 0;
+}
+
+.panel-content {
+  display: flex;
+  flex-direction: column;
+}
+
+.panel-content .setting-row {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  margin-bottom: 0.5rem;
+}
+
+.panel-content .setting-row:last-child {
+  margin-bottom: 0;
+}
--- a/src/Controls/settings-panel.jsx
+++ b/src/Controls/settings-panel.jsx
@@ -8,7 +8,6 @@ import 'react-modern-drawer/dist/index.css'
 import {
   Divider,
   Form,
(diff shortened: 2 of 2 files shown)


hs7898753:
Thankyou @MuneebAhmed25211  for you suggestions. 
But IMO having two drop down might also lead to confusion if some settings are present in both dropdowns.

MuneebAhmed25211:
Ok @hs7898753 I have another idea If we distribute the settings into different dropdowns under the settings dropdown e.g. I see there are 3 options of length, width and height in the extra settings options. If we combine these 3 into a single dropdown (e.g. dimensions) in the extra settings menu, it will take less space and will be clean. If we choose the names of the dropdowns properly and will make the pairs properly (like length, width, height). I suppose it will make no confusion according to me. Hope you understand.

hs7898753:
Yah! We can do it. thank you

nilsnolde:
> If we combine these 3 into a single dropdown (e.g. dimensions) in the extra settings menu, it will take less space and will be clean.

Yeah I like that idea! It's true, there's so much space wasted actually. Probably we'll need to even get rid of "General" & "Extra" settings and just provide _one_ settings which is unique for the selected mode. Grouping them like that was mostly due to not repeating code that much. But that really made the code much more confusing and spaghetti, maybe it's better to have the settings for each mode cleanly separated in the code as well at the expense of duplicating the same options a few times.

Then it needs a "careful" grouping of costing options which makes sense. The ones you mentioned make sense. There could also be one for e.g. "road classes" which groups `use_highways/living_streets/tracks` etc. That'd slim down the UI immensely if grouped right.

Good suggestion, thanks!

hs7898753:
Yah! I agree with the suggestion. 
What else I think is how is that if we have a icons for the different grouped settings. 
I think it will look good

nilsnolde:
If there are good icons for what the groups, sure.. I'd say one after the other, and first have a good outline of the grouping

If we go that way, it's a bigger job and more suitable for the GSoC project, rather than now. It fits into the whole re-write we plan to do. I'll transfer the idea to the appropriate issue.

PR #164, opened 2023-03-31
Improve styles for Isochrone tab

### LINKED ISSUE #162: [BUG] Current Isochrones tab enhancement
The current Isochrone tab looks like this. "Visualized with Valhalla App" at the end is getting cut off due to the width of the tab. Also the Directions tab and the Isochrone tab should be of same width, I think that would make more sense UX wise.

Keeping the width of Directions tab and Isochrone tab same. 

I would like to work on this issue, if possible.

### Screenshots

image

Sheikh-JamirAlam:
## 🛠️ Fixes Issue
Fixes #162 

## 👨‍💻 Changes proposed
- Made the input field vertically align with the delete button
- Made sure the footer is visible clearly.

## 📄 Note to reviewers
Added Screenshots

## 📷 Screenshots
BEFORE
image

AFTER
image

### DIFF
--- a/.env
+++ b/.env
@@ -6,4 +6,4 @@ REACT_APP_CENTER_COORDS="52.51831,13.393707"
 
 # uncomment this variable to set boundaries on the leaflet map
 #                       southwest corner, northeast corner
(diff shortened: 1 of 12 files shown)


nilsnolde (approved):
Thanks

Issue #172, opened 2023-04-04
Invalid size props are passed


rohan-kulkarni-25:
### What is the problem you are facing

When I deployed the project on my local system and was doing some random changes I noticed that in the console a warning is shown.

![Screenshot from 2023-04-04 15-36-55](https://user-images.githubusercontent.com/76584243/229761655-36e70bd0-2965-41fe-8cd0-d22c5f9b618d.png)

I just went through the code and found it in `Waypoint. jsx` size props are directly passed as `32px` This was the reason for the warning.

### What is your suggested solution

As we are following the proper syntax of `tiny` and `small` we should use the same variables here.
I have made the changes to check if is this only the reason for the error. 

![Screenshot from 2023-04-04 15-44-14](https://user-images.githubusercontent.com/76584243/229762140-b4bbb166-5ed9-4439-affc-62a2ecf5da24.png)

And yes the errors were gone. I hope this will help to maintain our code consistency, I am also submitting my proposal under this project in gsoc. Would love to make more such changes which will help us to maintain code consistency and improve our web app UI. 

Thanks

nilsnolde:
Thanks for this. GSoC is finished though, sorry. I mean you can apply still, but it's been too many good applicants already to be considered at this point.

rohan-kulkarni-25:
> Thanks for this. GSoC is finished though, sorry. I mean you can apply still, but it's been too many good applicants already to be considered at this point.

Ahh, no worries. Would love to apply still as I am interested to contribute on this project.

PR #173, opened 2023-04-04
passed the props in more consitent way

rohan-kulkarni-25:
## 👨‍💻 Changes proposed

- have passed the props in a more consistent way

## 📄 Note to reviewers

- changes are made in `Waypoint.jsx` file

## 📷 Screenshots
Have attached in issue #172


hs7898753:
well, I think this just reduced the size of the button.

Screenshot (2)

it will be nice if we use size="small" or "large".

Sheikh-JamirAlam:
@hs7898753 Yeah I noticed this as well. It's not very convinient.

PR #175, opened 2023-04-04
added vite

samyabrata-maji:
## 👨‍💻 Changes proposed
- [x] added vite as dev dependency
- [x] added scripts to run vite servers
- [x] created main.jsx which is the entrypoint (if vite is used). For now main.jsx has a dummy text.
- [x] added path aliases for vite


samyabrata-maji:
- [x] added path aliases for vite

samyabrata-maji:
Fixed parsing error
Added `@babel/eslint-parser` in the `.eslintrc.json` file

parser
parser2

samyabrata-maji:
1. Vite uses `import.meta.env` instead of `process.env` to access environment variables
2. Only the environment variables starting with `VITE_` are exposed.

vite.config.js
```diff
@@ -0,0 +1,34 @@
...
+const path = require('path')
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  resolve: {
+    alias: [
+      {
+        find: 'actions',
+        replacement: path.resolve(__dirname, './src/actions'),
+      },
```
nilsnolde, line 14: so it needs to be told how to resolve import paths? or what is that for?

samyabrata-maji: It matches for the word in the import statement and replaces it with the given path.

so, vite interprets: `import "action/file.jsx"` as  `import "./src/actions/file.jsx"`

Its optional.

PR #176, opened 2023-04-04
Clean Up Repo Issue solved #143

### LINKED ISSUE #143: [OTHER] Clean up repo

I noticed that the files are a bit unorganized/inconsistent in my opinion and I would like to clean it up **during the GSoC 2023** alongside #68. (Not the red, not sure why those are there, could be VSCode.)

Capture

I am suggesting to revise the files like so:

ComponentName - folder name
- ComponentName.jsx - the functional component 
- ComponentName.css - a css file to overwrite the parent css
- index.js - import ComponentName from ./ComponentName.jsx and export default ComponentName so files can just import as ./ComponentName instead of ./ComponentName/ComponentName
- other necessary js files/components - children components or js files used solely in ComponentName

This would make the repo easier to work with.

shreyarora12:
## 📄 Note to reviewers

Cleaner Code

### DIFF
--- a/src/index.css
+++ b/src/index.css
@@ -34,7 +34,7 @@ i.help.grey.icon {
 .custom-label {
   font-size: 0.85em;
   font-weight: bold;
-  color: #000000de;
+  color: rgba(0, 0, 0, 0.87);
 }
 
 .ui.toggle.checkbox input:focus:checked ~ .box:before,
--- a/src/index.css
+++ b/src/index.css
@@ -34,7 +34,7 @@ i.help.grey.icon {
 .custom-label {
   font-size: 0.85em;
   font-weight: bold;
-  color: rgba(0, 0, 0, 0.87);
+  color: #000000de;
 }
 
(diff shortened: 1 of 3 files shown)


src/index.css
```diff
@@ -34,7 +34,7 @@ i.help.grey.icon {
 .custom-label {
   font-size: 0.85em;
   font-weight: bold;
-  color: rgba(0, 0, 0, 0.87);
+  color: #000000de;
```
nilsnolde, line 37: I'd like to revert this one, I find it much more legible to see the opacity in float rather than hex

nilsnolde (changes requested):
This technically doesn't solve #143 entirely, but it does do a tiny bit of cleanup. I'd only ask to please revert the hex color change.

shreyarora12:
I came across the project sort of a bit late, that's why I missed it. I
wanted to submit a proposal for the purpose of GSOC but since I was short
on time I couldn’t. I’ll correct the issue right away.

PR #184, opened 2023-04-12
ci: trigger preview for PR

Ananya2001-an:
## 👨‍💻 Changes proposed
Triggers a preview when a PR is raised. 

## 📄 Note to reviewers
Ready for review

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -10,15 +10,18 @@ on:
       - 'jsconfig.json'
       
 env:
-  PULL_NUMBER: $(jq --raw-output .pull_request.number "$GITHUB_EVENT_PATH")
+  PULL_NUMBER: ${{ github.event.pull_request.number }}
 
 jobs:
   preview:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: update homepage URL
-        run: echo '{"homepage":"https://valhalla-app-tests.gis-ops.com/'$PULL_NUMBER'"}' > package.json
+      - name: 🌐 Update homepage URL
+        run: |
+          NEW_HOMEPAGE="https://valhalla.openstreetmap.de/$PULL_NUMBER"
+          jq --arg homepage "$NEW_HOMEPAGE" '. + {homepage: $homepage}' package.json > tmp.json
+          mv tmp.json package.json
       - uses: actions/setup-node@v2
         with:
           node-version: '16.x'
@@ -30,4 +33,4 @@ jobs:
         run: |
           eval $(ssh-agent -s)
           ssh-add - <<< "${{ secrets.SSH_PRIVATE_KEY }}"
-          rsync -r --delete --mkdir -e "ssh -p 23432 -o StrictHostKeyChecking=no" ./build/ ${{ vars.HOST_USERNAME }}@${{ vars.HOST_IP }}:${{ vars.HOST_TARGET_DIR }}/$PULL_NUMBER
+          rsync -r --delete --rsync-path="mkdir -p ${{ vars.HOST_TARGET_DIR }}/$PULL_NUMBER && rsync" -e "ssh -p 23432 -o StrictHostKeyChecking=no" ./build/ ${{ vars.HOST_USERNAME }}@${{ vars.HOST_IP }}:${{ vars.HOST_TARGET_DIR }}/$PULL_NUMBER
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -2,26 +2,34 @@ name: PR preview
 
 on:
   pull_request:
+    types: [opened, reopened, synchronize, closed]
     paths:
       - 'src/**'
       - 'public/**'
       - 'package*.json'
       - '.env'
       - 'jsconfig.json'
-      
+
 env:
   PULL_NUMBER: ${{ github.event.pull_request.number }}
 
 jobs:
-  preview:
+  clean-up:
     runs-on: ubuntu-latest
+    if: ${{ github.event.action == 'closed' }}
+    steps:
+      - name: 🧹 Remove preview dir
+        run: |
+          eval $(ssh-agent -s)
+          ssh-add - <<< "${{ secrets.PREVIEW_SSH_PRIVATE_KEY }}"
+          ssh -o StrictHostKeyChecking=no ${{ vars.PREVIEW_HOST_USERNAME }}@${{ vars.PREVIEW_HOST_IP }} "rm -rf ${{ vars.PREVIEW_HOST_TARGET_DIR }}/$PULL_NUMBER"
+  deploy:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.action != 'closed' }}
     steps:
       - uses: actions/checkout@v2
-      - name: 🌐 Update homepage URL
-        run: |
-          NEW_HOMEPAGE="https://valhalla.openstreetmap.de/$PULL_NUMBER"
-          jq --arg homepage "$NEW_HOMEPAGE" '. + {homepage: $homepage}' package.json > tmp.json
-          mv tmp.json package.json
+      - name: 🏠 Replace homepage URL
+        run: sed -i "s/https:\/\/valhalla.openstreetmap.de/https:\/\/valhalla-app-tests.gis-ops.com\/${PULL_NUMBER}/g" package.json
       - uses: actions/setup-node@v2
         with:
           node-version: '16.x'
@@ -29,8 +37,21 @@ jobs:
         run: |
           npm install
           npm run build
-      - name: 🚀 Preview deployment
+      - name: 🚀 Preview deploy
         run: |
           eval $(ssh-agent -s)
-          ssh-add - <<< "${{ secrets.SSH_PRIVATE_KEY }}"
-          rsync -r --delete --rsync-path="mkdir -p ${{ vars.HOST_TARGET_DIR }}/$PULL_NUMBER && rsync" -e "ssh -p 23432 -o StrictHostKeyChecking=no" ./build/ ${{ vars.HOST_USERNAME }}@${{ vars.HOST_IP }}:${{ vars.HOST_TARGET_DIR }}/$PULL_NUMBER
+           ssh-add - <<< "${{ secrets.PREVIEW_SSH_PRIVATE_KEY }}"
+           rsync -r --delete --rsync-path="mkdir -p ${{ vars.PREVIEW_HOST_TARGET_DIR }}/$PULL_NUMBER && rsync" -e "ssh -o StrictHostKeyChecking=no" ./build/ ${{ vars.PREVIEW_HOST_USERNAME }}@${{ vars.PREVIEW_HOST_IP }}:${{ vars.PREVIEW_HOST_TARGET_DIR }}/$PULL_NUMBER
+      - name: 🌐 Preview URL
+        if: success()
+        uses: actions/github-script@v4
+        with:
+          github-token: ${{ secrets.PAT }}
+          script: |
+            const body = `Preview is ready! 🚀 You can view it here: https://valhalla-app-tests.gis-ops.com/${{ github.event.pull_request.number }}`;
+            const { data } = await github.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              body: body
+            });
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -1,0 +1,33 @@
+name: PR preview
+
+on:
+  pull_request:
+    paths:
+      - 'src/**'
+      - 'public/**'
+      - 'package*.json'
+      - '.env'
+      - 'jsconfig.json'
+      
+env:
+  PULL_NUMBER: $(jq --raw-output .pull_request.number "$GITHUB_EVENT_PATH")
+
+jobs:
+  preview:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: update homepage URL
+        run: echo '{"homepage":"https://valhalla-app-tests.gis-ops.com/'$PULL_NUMBER'"}' > package.json
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16.x'
+      - name: 🔧 Create build
+        run: |
+          npm install
+          npm run build
+      - name: 🚀 Preview deployment
+        run: |
+          eval $(ssh-agent -s)
+          ssh-add - <<< "${{ secrets.SSH_PRIVATE_KEY }}"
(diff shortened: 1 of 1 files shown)


Ananya2001-an:
Hey @nilsnolde I have created a draft PR for feedback purpose.

Ananya2001-an:
Also right now the build is failing saying that the script `build` is missing. You can check here: https://github.com/Ananya2001-an/valhalla-app/actions/runs/4678214265/jobs/8286676179?pr=3

Ananya2001-an:
I have not made all changes yet so don't test till I tell u to.

.github/workflows/preview.yml
```diff
@@ -0,0 +1,33 @@
...
+env:
+  PULL_NUMBER: $(jq --raw-output .pull_request.number "$GITHUB_EVENT_PATH")
+
+jobs:
+  preview:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: update homepage URL
+        run: echo '{"homepage":"https://valhalla-app-tests.gis-ops.com/'$PULL_NUMBER'"}' > package.json
```
nilsnolde, line 21: did you try this locally? it'll result in a broken package.json. you'll need to replace only the value of "homepage", not the entire file.

Ananya2001-an:
It's working now you can see this preview for the PR on my fork which changes the title of the app: https://valhalla-app-tests.gis-ops.com/3/

Ananya2001-an:
Now I just have to add a step for deletion of the folder on PR close. Also I forgot to mention earlier but we have to update that permalink function as well....

Ananya2001-an:
> It's not essential for this to work, but it's still weird code yeah.

Yeah but like if someone does changes that involves making API calls then that would cause trouble.

.github/workflows/preview.yml
```diff
@@ -0,0 +1,36 @@
...
+  PULL_NUMBER: ${{ github.event.pull_request.number }}
+
+jobs:
+  preview:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: 🌐 Update homepage URL
+        run: |
+          NEW_HOMEPAGE="https://valhalla.openstreetmap.de/$PULL_NUMBER"
```
nilsnolde, line 22: shouldn't this be https://valhalla-app-tests.gis-ops.com/$PULL_NUMBER?

.github/workflows/preview.yml
```diff
@@ -0,0 +1,36 @@
...
+          jq --arg homepage "$NEW_HOMEPAGE" '. + {homepage: $homepage}' package.json > tmp.json
+          mv tmp.json package.json
```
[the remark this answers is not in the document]
Ananya2001-an: Let me test it. This looks good tho 👍🏼

Ananya2001-an:
Everything is done now:

- [x] Added clean up job for closed/merged PRs
- [x] Also updated the homepage url step as u suggested
- [x] Also added a step where as soon as deployment would be done successfully, a comment on the PR will be sent with the preview URL. 

We just need to create the necessary vars and secrets. Also for the last step `Preview URL` we would need a Personal access token to use github API and make the POST request. I have changed the variable and secret names since this host server is different from the original one. Hope everything is good :) 


Also u can see this PR on my fork to see the kind of comment we are getting on successful deployment: https://github.com/Ananya2001-an/valhalla-app/pull/4

.github/workflows/preview.yml
```diff
@@ -0,0 +1,57 @@
+name: PR preview
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, closed]
```
nilsnolde, line 5: I'm kinda wondering if we shouldn't only do this on request, i.e. with `workflow_dispatch`? WDYT @Ananya2001-an @chrstnbwnkl ? 

It doesn't take that long and is tolerable to wait for when a reviewer wants to preview. I at least, usually first review the code before I try the app itself, unless it's only CSS changes or so, but happens rarely.

I'm not so sure though how to conditionally run a step only if it's coming from a `workflow_dispatch` event, but I'm sure that's possible, it should be somewhere in the context of Github, I think I saw it earlier, but didn't mention it, wanted to see what you come up with:)

.github/workflows/preview.yml
```diff
@@ -0,0 +1,57 @@
...
+
+jobs:
+  clean-up:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.action == 'closed' }}
+    steps:
+      - name: 🧹 Remove preview dir
+        run: |
+          eval $(ssh-agent -s)
+          ssh-add - <<< "${{ secrets.PREVIEW_SSH_PRIVATE_KEY }}"
```
nilsnolde, line 24: good thinking, better not to use the same ssh key as for deployment

.github/workflows/preview.yml
```diff
@@ -0,0 +1,57 @@
...
+          ssh -o StrictHostKeyChecking=no ${{ vars.PREVIEW_HOST_USERNAME }}@${{ vars.PREVIEW_HOST_IP }} "rm -rf ${{ vars.PREVIEW_HOST_TARGET_DIR }}/$PULL_NUMBER"
+  deploy:
```
nilsnolde, line 26: ah I actually remember now that I used multiple jobs in other workflows and I remember that you'll need to define dependencies, i.e. this workflow shouldn't run before the `clean-up` workflow. I'm not sure though: do they run in parallel if you don't set a `needs: 'clean-up` on the `deploy` job? if they do run in parallel then we should be fine, since the build takes so much longer than the tiny ssh command on `clean-up`

.github/workflows/preview.yml
```diff
@@ -0,0 +1,57 @@
...
+    runs-on: ubuntu-latest
+    if: ${{ github.event.action != 'closed' }}
+    steps:
+      - uses: actions/checkout@v2
```
nilsnolde, line 30: I saw today this action is at `v3` by now

.github/workflows/preview.yml
```diff
@@ -0,0 +1,57 @@
...
+          npm install
+          npm run build
+      - name: 🚀 Preview deploy
+        run: |
+          eval $(ssh-agent -s)
+           ssh-add - <<< "${{ secrets.PREVIEW_SSH_PRIVATE_KEY }}"
+           rsync -r --delete --rsync-path="mkdir -p ${{ vars.PREVIEW_HOST_TARGET_DIR }}/$PULL_NUMBER && rsync" -e "ssh -o StrictHostKeyChecking=no" ./build/ ${{ vars.PREVIEW_HOST_USERNAME }}@${{ vars.PREVIEW_HOST_IP }}:${{ vars.PREVIEW_HOST_TARGET_DIR }}/$PULL_NUMBER
+      - name: 🌐 Preview URL
+        if: success()
+        uses: actions/github-script@v4
```
nilsnolde, line 47: seems like we should use `v6`, it's using node 16: https://github.com/actions/github-script#breaking-changes-in-v6

.github/workflows/preview.yml
```diff
@@ -0,0 +1,57 @@
...
+        with:
+          github-token: ${{ secrets.PAT }}
```
nilsnolde, line 49: what's this? it sounds like we don't need a custom token: https://github.com/actions/github-script#using-a-separate-github-token

.github/workflows/preview.yml
```diff
@@ -0,0 +1,57 @@
...
+          script: |
+            const body = `Preview is ready! 🚀 You can view it here: https://valhalla-app-tests.gis-ops.com/${{ github.event.pull_request.number }}`;
+            const { data } = await github.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              body: body
+            });
```
nilsnolde, line 57: this would be pretty cool:) 

but how this work? it seems to try to create a comment in an _issue_? I can't see how that works on a _pull request_? I guess a typo?

also, in case we'd go with a `workflow_dispatch` I think we could still use smth like this. it might be that we only get the branch _name_ though, not the number. which would slightly suck (it's not unique!), but would IMO almost be worth the convenience of an automated comment that gives me the link right there.

.github/workflows/preview.yml
nilsnolde: our does the default `github-token` not have access to that ressource? probably not.. right, then we need a custom token with rights to write comments to an issue. I'll create one!

nilsnolde:
> Also u can see this PR on my fork to see the kind of comment we are getting on successful deployment: https://github.com/Ananya2001-an/valhalla-app/pull/4

sorry, just reading your comment. huh, strange, I was wondering how would the preview comment work if there's `.issue`, not `.pull_request` referenced in the API request. but then this does create a comment in this PR, not in an issue, right? weird..

I created an access token for this and registered all vars & secrets. at least I think I did the right thing with the access token, it became pretty confusing since the last time I did that..

I wonder why this workflow wasn't triggered on your last push? it was a `synchronize` event and checked all the other boxes no?

nilsnolde:
> I wonder why this workflow wasn't triggered on your last push? it was a synchronize event and checked all the other boxes no?

ah it didn't because it's not part of the files it's watching out for. we should probably include the respective workflow file as well: we'll want to run the workflow when we change smth.

.github/workflows/preview.yml
nilsnolde: if possible, we could use the owner of the access token: `github-bot` as the commenter. but maybe that'll happen automatically because the personal access token is registered by that user?

nilsnolde:
One other thing we could add: running the linters before building & deploying the app. I'm not sure what we use in this repo tbh, at least `prettier` I guess? others?

IMO best to include that as command in package.json so we can just call it with npm.

nilsnolde:
apparently `eslint` as well? probably that also needs some `==` fixes and similar.. not sure how sound our eslint config is  though..

nilsnolde:
actually I'd almost prefer to run this only on `workflow_dispatch`. I can imagine how this clutters the PR interface terribly, if there's one comment for each commit. but even for manual dispatches it's still potentially very noisy, and worse, without any added value, the URL will never change for a PR. 

hmm.. since there's a script running, one could in theory check all comments on this PR and there's any comment by the user `gisops-bot` we don't create an issue? if there's _any_ comment by that user on a PR, we know that we already have the comment with the URL. opinions?

nilsnolde:
sorry for being so nitpicky, but this is a really good action which we'll use in many other (at least private) projects I think.

.github/workflows/preview.yml
nilsnolde: I'm also perfectly happy to leave it as-is. as long as we make sure we only add a comment once, see https://github.com/gis-ops/valhalla-app/pull/184#issuecomment-1505994154

Ananya2001-an: Yeah I will do something to add that comment only once...makes sense since it's the same url

Ananya2001-an: And about workflow dispatch I don't think that's possible since it depends on the Pull request and on triggering it wont know which one we are talking about.

.github/workflows/preview.yml
```diff
@@ -0,0 +1,57 @@
...
+      - 'public/**'
+      - 'package*.json'
+      - '.env'
+      - 'jsconfig.json'
+
+env:
+  PULL_NUMBER: ${{ github.event.pull_request.number }}
+
+jobs:
+  clean-up:
```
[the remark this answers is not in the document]
Ananya2001-an: Yeah true 👍🏼

.github/workflows/preview.yml
Ananya2001-an: Since we have the if condition defined in the `deploy` job, it will not run when the PR will be closed. Yes, they run parallel but before executing the steps by looking at the if condition it would skip that job depending on what PR action was executed (opened, closed....).

.github/workflows/preview.yml
Ananya2001-an: Okay need to update the version then..

.github/workflows/preview.yml
```diff
@@ -0,0 +1,57 @@
...
+          ssh-add - <<< "${{ secrets.PREVIEW_SSH_PRIVATE_KEY }}"
+          ssh -o StrictHostKeyChecking=no ${{ vars.PREVIEW_HOST_USERNAME }}@${{ vars.PREVIEW_HOST_IP }} "rm -rf ${{ vars.PREVIEW_HOST_TARGET_DIR }}/$PULL_NUMBER"
+  deploy:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.action != 'closed' }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: 🏠 Replace homepage URL
+        run: sed -i "s/https:\/\/valhalla.openstreetmap.de/https:\/\/valhalla-app-tests.gis-ops.com\/${PULL_NUMBER}/g" package.json
+      - uses: actions/setup-node@v2
```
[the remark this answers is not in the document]
Ananya2001-an: Sure 👍🏼

.github/workflows/preview.yml
Ananya2001-an: No right it's going to show as if u wrote the comment since you are the repo's owner but u are right to say that the github bot should be the owner I will look into it.

Ananya2001-an:
> sorry, just reading your comment. huh, strange, I was wondering how would the preview comment work if there's `.issue`, not `.pull_request` referenced in the API request. but then this does create a comment in this PR, not in an issue, right? weird..

Actually every PR is also an issue in GitHub and what actually matters is the number so this same api would work to create a comment inside the issue as well as a PR.

> I created an access token for this and registered all vars & secrets. at least I think I did the right thing with the access token, it became pretty confusing since the last time I did that..

Great 👍🏼

Ananya2001-an:
> hmm.. since there's a script running, one could in theory check all comments on this PR and there's any comment by the user `gisops-bot` we don't create an issue? if there's _any_ comment by that user on a PR, we know that we already have the comment with the URL. opinions?


Yeah that's what I would look into

Ananya2001-an:
> sorry for being so nitpicky, but this is a really good action which we'll use in many other (at least private) projects I think.

No worries.. I appreciate detailed feedback and definitely this will be a good action to use everywhere else so we need to make sure that it's done right the first time :)

Ananya2001-an:
> Jep just one more tiny thing: we’ll also need to add the PR workflow file to the list of files triggering a build. That’ll also tell if it worked before merging.

Yeah u are right we will do that.

> You’re right, workflow_dispatch wouldn’t get the PR, only the branch. We could do smth more elaborate with the script and GitHub API but it’s not worth the effort.
 
Exactly 👍🏼

nilsnolde:
Jep, I'll look at it in a bit

nilsnolde:
Woohoo, it worked! This is great @Ananya2001-an :) Seems like the only thing we need to watch out for, is that we don't immediately merge after triggering a build. I'd imagine that'd first remove the folder on our server and then re-create it. Small price to pay and happens rarely.

nilsnolde (approved):

PR #188, opened 2023-06-09
Fix double call to route API on truck profile

jthiard:
Hi, thanks for this very useful Valhalla web frontend!

## 👨‍💻 Changes proposed

I noticed duplicate calls to route API  from / to the truck profile.
I guessed this only happens with the truck profile because it is the only one one having a different set of default settings.

## 📄 Note to reviewers

I'm not familiar with all the effects / actions in this codebase, so I'm not sure this is the right change, but it fix the issue.

## 📷 Screenshots

Relevant screenshot from https://valhalla.openstreetmap.de/
The first API call is made with the car profile.
I then switched to the truck profile, and two API calls were fired at same time, one of them getting a 429 error (too many requests).

Screenshot_2023-06-09_11-42-46

### DIFF
--- a/src/Controls/settings-panel.jsx
+++ b/src/Controls/settings-panel.jsx
@@ -130,7 +130,10 @@ class SettingsPanel extends React.Component {
       this.setState({ generalSettings, extraSettings })
     }
 
-    if (!R.equals(settings, nextProps.settings)) {
+    if (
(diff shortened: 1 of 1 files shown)


nilsnolde:
hmpf, smth changed in our secrets or so.. no time right now, but this looks good. thanks!

PR #189, opened 2023-06-20
Switch to tile.openstreetmap.org

otbutz:
## 👨‍💻 Changes proposed

See https://github.com/openstreetmap/operations/issues/737


nilsnolde:
Oh sweet thanks a lot! Didn't realize they changed that. Also TIL there's HTTP/3 :smile:

nilsnolde:
Never mind CI failure, some vars wrongly set.

PR #196, opened 2023-09-22
motorcycle profile for direction (copy from car)

nvcnvn:
## 👨‍💻 Changes proposed

Adding motorcycle profile (for now just copy everything from auto)

## 📄 Note to reviewers

I don't have a proper motorcycle icon, so I reuse bike icon

## 📷 Screenshots
Screenshot 2023-09-22 at 16 40 39


nilsnolde (changes requested):
Copying (mostly) from auto should be fine, but we'd want
- at least `use_trails`: https://valhalla.github.io/valhalla/api/turn-by-turn/api-reference/#motorcycle-costing-options-beta
- definitely a motorcycle SVG

before merging.

src/Controls/settings-options.js
```diff
@@ -688,6 +700,7 @@ export const settingsInit = {
   transit_start_end_max_distance: 2145,
   transit_transfer_max_distance: 800,
   disable_hierarchy_pruning: false,
+  use_trails: 1,
```
nilsnolde, line 703: the default needs to be 0 as per docs

nilsnolde (changes requested):
Just one more minor thing, then it's good to merge.

src/Controls/settings-options.js
nvcnvn: ooppps, not sure why I missed that.

nilsnolde:
pity the PR workflow is broken, would love to see it before merging. not sure what's wrong there. seems the secret is not really taken by actions. I just re-pasted it but  still nothing.. smth must've changed in how secrets are referred to.

anyways, thanks for the changes @nvcnvn , looks safe to me to deploy regardless.

PR #200, opened 2023-10-25
add date-time option for easy testing conditional-turn-restriction

nvcnvn:
## 👨‍💻 Changes proposed
- Add date-time picker (for easier testing time-based turn restriction)

## 📷 Screenshots
Screenshot 2023-10-25 at 15 37 01


nilsnolde:
Thanks, that'd be a nice addition which would make the app much more useful. A couple of comments:
- seems like it's now mandatory to give a date/time? Would be better if it was opt-in, e.g. the default item in the dropdown could be "Anytime" or so and a greyed out date/time field. WDYT?
- the `moment` library causes a build failure: https://github.com/gis-ops/valhalla-app/actions/runs/6637880358/job/18036721164?pr=200. Can you have a look?

nvcnvn:
> * seems like it's now mandatory to give a date/time? Would be better if it was opt-in, e.g. the default item in the dropdown could be "Anytime" or so and a greyed out date/time field. WDYT?

I just want to clarify the behavior of Valhalla in this case, should "Anytime" (empty date_time)  different with "Current" (date_time.type = 0)?
Because most of the time we: A don't care about time, B want to test departure/arrival.

nilsnolde:
Yes it will be different. If you say datetime.type = 0, then it’ll use the current time, ie time will be respected (incl restrictions, but also traffic if it’s loaded). When you don’t want to respect time, then you don’t send the date_time object in the request at all. And I agree, the latter will be the vast majority of requests.

src/components/datetime-picker.jsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react'
```
chrstnbwnkl, line 1: File name should be pascal case

src/components/datetime-picker.jsx
```diff
@@ -0,0 +1,62 @@
...
+import PropTypes from 'prop-types'
+import { Input, Dropdown } from 'semantic-ui-react'
+
+export const DateTimePicker = ({ type, value, onChange }) => {
+  return (
+    <div
+      className="pa2 flex flex-wrap justify-between"
+      style={{
+        marginLeft: 'auto',
```
chrstnbwnkl, line 10: doesn't seem to have any effect

src/Controls/Directions/index.jsx
```diff
@@ -62,8 +68,24 @@ class DirectionsControl extends React.Component {
...
+      date_time.value !== nextProps.date_time.value
+    if (shouldUpdate) {
+      this.props.dispatch(makeRequest())
+    }
+    return shouldUpdate
+  }
+
   render() {
-    const { profile, loading } = this.props
+    const { profile, loading, date_time } = this.props
```
chrstnbwnkl, line 88: again, please use camelCase

chrstnbwnkl (changes requested):
Thanks @nvcnvn , just a couple of comments!

nvcnvn:
> Thanks @nvcnvn , just a couple of comments!

Finally I found some time to update the PR.

chrstnbwnkl (changes requested):
Sorry for not getting back to this earlier... Looks like you forgot to check in the renamed datepicker component file?

nvcnvn:
> Sorry for not getting back to this earlier... Looks like you forgot to check in the renamed datepicker component file?

ah you are right

chrstnbwnkl (approved):
...and again sorry for the long wait. LGTM!

PR #213, opened 2024-02-15
Improve isochrones

chrstnbwnkl:
Added denoise and generalize settings for isochrones, added a isochrone location layer for the snapped locations, and prepared for upcoming multipolygon support (without breaking current polygons, at least not more than it was doing anyway :sweat_smile: denoise was not set at all, so any rings are pretty much ignored anyway currently).

### DIFF
--- a/src/Controls/Isochrones/ContoursInformation.jsx
+++ b/src/Controls/Isochrones/ContoursInformation.jsx
@@ -26,30 +26,34 @@ class ContoursInformation extends React.Component {
     return (
       <React.Fragment>
         {features ? (
-          features.map((feature, key) => {
-            return (
-              <div className={'flex pb2'} key={key}>
-                <div
-                  className={'flex'}
-                  style={{
-                    alignSelf: 'center',
-                    flexBasis: '140px',
-                  }}
-                >
-                  <Icon circular name={'time'} />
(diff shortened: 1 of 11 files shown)


chrstnbwnkl:
image

Sweet, looks like Leaflet is *very* lenient when it comes to rendering polygon rings :smile:

nilsnolde:
Huh, did you change smth in the secrets? Previously there was a very obnoxious error where it couldn't read the secret for some reason and I've not been able to fix that in a medium-effort attempt. I supposed it was a Github bug, awesome that it works again!

nilsnolde:
Only thing I'd change is the location of the parameters. I think that's smth that deserves to live in the main isochrone interface on the left side. It's very very sensitive, 0.1 removes all of those holes you have up there. What do you think?

Anyways, thanks for the effort already, I hope people will appreciate it too. Now isochrones have MUCH better support.

nilsnolde:
Nit (only if it's not a bunch of work):
image

chrstnbwnkl:
No changes in the env, not sure why how or how that issue got resolved :smile:

nilsnolde (approved):
> at least they're URL synced now!

that's indeed good, esp for issue reproduction!

I'm not gonna look at all at the code :see_no_evil: :laughing:

PR #217, opened 2024-04-11
Expose ignore_access and ignore_restrictions API flags

eikes:
## 👨‍💻 Changes proposed

Expose ignore_access and ignore_restrictions API flags to show how they change the generated routes

### DIFF
--- a/src/Controls/settings-options.js
+++ b/src/Controls/settings-options.js
@@ -944,7 +944,7 @@ export const settings_general = {
       ignoreClosures,
       ignoreRestrictions,
       ignoreAccess,
-      excludeCashOnlyTolls
+      excludeCashOnlyTolls,
     ],
     enum: [],
   },
@@ -967,7 +967,7 @@ export const settings_general = {
       ignoreClosures,
       ignoreRestrictions,
       ignoreAccess, 
-      excludeCashOnlyTolls
+      excludeCashOnlyTolls,
     ],
     enum: [],
   },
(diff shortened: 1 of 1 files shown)


nilsnolde:
Thanks @eikes! eslint failed it seems. once that’s fixed we’ll get a preview.

nilsnolde:
there's now other eslint problems. does a `npm install && npm run build` work for you? it doesn't seem like CI executes eslint explicitly: https://github.com/gis-ops/valhalla-app/blob/2661dc687005a3e75b908d78a81eb19b2dbe8d09/.github/workflows/deploy.yml#L26C11-L26C24. so I guess it's part of `npm run build`.. (not so fond of and used to JS ;))

eikes:
I totally misread the eslint / prettier output. I thought it wanted me to remove trailing commas, but it was showing a symbol for space, which I have finally found and removed:

```
Delete `·`  prettier/prettier
```

eikes:
The "Preview Deploy" step still seems to have failed. But I don't think that's related to these changes.

nilsnolde:
sorry for the long wait! yeah right, I forgot that forks don't have access to our secrets, sorry about that.. that makes our workflow really less useful. github should just require an admin to approve running the workflow if it's coming from a fork and ideally even tell me that "attention, secret being used in workflow". I don't know how to set this up better in the meantime. gh pages won't work, no dynamic subdomain possible. it's such a nice idea, but half-fails in reality..

eikes:
Looks good on valhalla.openstreetmap.de now. Thanks!

PR #238, opened 2025-08-15
Remove duplicate settings

### LINKED ISSUE #235: Duplicate settings border cost and border penalty
For some costing types (car, truck, bus, motorcycle) the settings border cost and border penalty appear twice. Once in Extra Settings and once in General Settings.

Have these settings appear only once. 
I'm unsure how these are categorized into extra and general, so not sure where the correct place would be.

### Screenshots

Image

tristancorbellari:
## 🛠️ Fixes Issue

Closes #235

## 👨‍💻 Changes proposed

- Removed the "Border Cost" and "Border Penalty" settings from duplication in "General Settings", so that they only live in "Extra Settings" now. If "General Settings" is a better place for them to live, I'll be happy to move them there instead.
- Corrected some inconsistent capitalisation and spelling mistakes ("penality" 🤪)

### DIFF
--- a/src/Controls/settings-options.js
+++ b/src/Controls/settings-options.js
@@ -929,8 +929,6 @@ export const settings_general = {
   truck: {
     numeric: [
       turnPenalityCost,
-      countryCrossingPenality,
-      countryCrossingCost,
       useHighways,
       useTollways,
       tollBoothCost,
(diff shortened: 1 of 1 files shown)


tristancorbellari:
I don't have secrets set up on this fork which I believe is why the workflow is failing

nilsnolde:
yeah sadly.. apparently it's possible: https://michaelheap.com/access-secrets-from-forks/. I'll give that a shot sometime. thanks for the cleanup!

PR #240, opened 2025-08-25
Fix Valhalla/Nominatim race condition

### LINKED ISSUE #227: Saved (Cut & Paste) route fails to redisplay after Pasteon>
### Description

https://valhalla.openstreetmap.de/directions?profile=car&wps=6.0972029%2C52.2093817%2C7.0408738%2C52.3124012%2C7.0679909%2C52.3202517%2C12.7509051%2C52.9060194%2C12.8532636%2C52.7535156

When I cut this route from Edge, and paste it into Firefox, all points are present, but on occasion the route does not go via all of them. Sometimes a second Paste will resolve the problem, but on other occasions the only way to get the route to be traced is to move the missing point a miniscule amount.

The issue not only happens with this route, but with dozens of others...

### Screenshots

_No response_

tristancorbellari:
## 🛠️ Fixes Issue

(Partially?) closes #227 

## 👨‍💻 Changes proposed

As seen in the issue above, there are sometimes cases when pasting a route URL results in some of the waypoints not being included in the route (though they are still present on the map). See below screenshots for example.

I managed to track this down to a race condition where the retrieval of the route directions from the Valhalla API would execute before all the waypoints' geocode results would be returned from the Nominatim API. This is why the issue would usually not arise when using the UI normally to add points one at a time, but instead occurs when pasting a route URL where multiple API calls now suddenly have to be executed all at once.

The nature of the issue means it is a little hard to reliably reproduce, but you should be able to encounter the issue when pasting URLs for routes with a good few waypoints (and maybe a couple refreshes as this behaviour is not always consistent).

My proposed change is to replace the previous logic for executing the Valhalla API call (using the `permaLast` variable) with a new solution that only executes once `pendingGeocodes` (the number of geocode results yet to be returned from the Nominatim API) is 0.

## 📄 Note to reviewers

I have experienced consistent correct behaviour in my testing of this new solution, though as this change concerns a fairly critical part of the application, I would encourage others to pull this branch and test it out for themselves, just to be sure!

## 📷 Screenshots

**Before:**

https://valhalla.openstreetmap.de/directions?profile=car&wps=6.0972029%2C52.2093817%2C7.0408738%2C52.3124012%2C7.0679909%2C52.3202517%2C12.7509051%2C52.9060194%2C12.8532636%2C52.7535156

When accessing the above URL, some points will sometimes not be included in the route. Here is a particularly bad example I was able to reproduce where only the start and end points (1 and 5) were included in the route and the rest were ignored:

image4
image5
image6

**After:**

Fixed, accessing the URL now consistently results in all points being included in the route:

image7
image8
image9

### DIFF
--- a/src/Controls/index.jsx
+++ b/src/Controls/index.jsx
@@ -95,7 +95,6 @@ class MainControl extends React.Component {
         const payload = {
           latLng,
           fromPerma: true,
-          permaLast: i === coordinates.length / 2 - 1,
           index: i,
         }
         processedCoords.push([latLng.lat, latLng.lng])
--- a/src/Controls/index.jsx
+++ b/src/Controls/index.jsx
@@ -94,7 +94,6 @@ class MainControl extends React.Component {
         const latLng = { lat: next, lng: current }
         const payload = {
           latLng,
-          fromPerma: true,
           index: i,
         }
         processedCoords.push([latLng.lat, latLng.lng])
--- a/src/actions/directionsActions.js
+++ b/src/actions/directionsActions.js
@@ -158,24 +158,15 @@ export const fetchReverseGeocodePerma = (object) => (dispatch) => {
   dispatch(requestGeocodeResults({ index: object.index, reverse: true }))
 
   const { index } = object
-  const { permaLast } = object
   const { lng, lat } = object.latLng
 
   if (index > 1) {
-    dispatch(doAddWaypoint(true, permaLast))
+    dispatch(doAddWaypoint(true))
   }
 
   reverse_geocode(lng, lat)
     .then((response) => {
-      dispatch(
-        processGeocodeResponse(
-          response.data,
-          index,
-          true,
-          [lng, lat],
-          permaLast
-        )
-      )
+      dispatch(processGeocodeResponse(response.data, index, true, [lng, lat]))
     })
     .catch((error) => {
       console.log(error) //eslint-disable-line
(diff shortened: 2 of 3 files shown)


nilsnolde:
@tristancorbellari sorry for the late answer, I just got back from a longer tour. As discussed offline, I added you as a contributor. 

Would you mind to close this PR and open a new one from the `valhalla` remote? Then our [GHA magic](https://github.com/valhalla/web-app/blob/master/.github/workflows/preview.yml) kicks in and builds the web app for us, publishes it and leaves a comment with an access link, so everyone can try this without building it locally. Other remotes don't have access to the repo's secrets (SSH key of the server used in this case).

nilsnolde:
Or never mind actually. Your solution makes sense to me! 

In case you'd like to contribute further I'd still recommend basing the PR on the upstream remote so it's easier for us to review:)

nilsnolde (approved):

tristancorbellari:
Thanks for bringing this to attention @mustaphaturhan, you're right that the routing does not seem to execute anymore when using the map's right-click controls. In the meantime, you can still use the address search feature while I get a fix in as soon as I can. My apologies!

nilsnolde:
Oops. I blame the non working GHA and the non-existence of tests. 

I actually didn't notice that, but what I did notice is that once there's a route, moving any waypoint also doesn't trigger a request.

I'll revert on the server to the previous commit.

PR #243, opened 2025-09-16
Remove jquery-ui-bundle and use vanilla js instead

### LINKED ISSUE #181: [error] jQuery is not defined
I am getting an error "ReferenceError: jQuery is not defined" while migrating to vite

As far as I can remember, it worked fine with `react-scripts`.

I have tried installing jQuery using yarn and using the cdn. Both did not work.

### Screenshots

Screenshot 2023-04-10 150549

mustaphaturhan:
Closes #181 

## 🛠️ Fixes Issue

While we want migrate the project to vite and modern development stack, one of the blocker was jQuery. To handle jQuery in vite, we have to use plugins like `rollup-plugin-inject`. But why would we need what when we can solve the problem at its root?

This PR closes #181 and removes the jQuery.

## 👨‍💻 Changes proposed

- Removes `jquery-ui-bundle` and `jQuery`
- Updates the `parserOptions` of `eslint` since we are using modern features of ecmascript.

## 📄 Note to reviewers

- I updated to `"ecmaVersion": 2022` because we are using `**?.**` syntax in the project.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -13,7 +13,7 @@
   "extends": ["eslint:recommended", "plugin:react/recommended"],
   "overrides": [],
   "parserOptions": {
-    "ecmaVersion": 9,
+    "ecmaVersion": 2022,
     "sourceType": "module",
     "ecmaFeatures": {
       "jsx": true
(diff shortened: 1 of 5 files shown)


nilsnolde:
Definitely, I'll build and run it later today. I don't have much to add on logical review, I'll just do some UX playing around.

PR #244, opened 2025-09-16
Migrate to vite

### LINKED ISSUE #171: [Suggestion] Migrate to Vite
As you know, CRA is slow. I recommend using Vite. Even the official react docs recommend using vite.

Using Vite would significantly reduce build time, improving developer experience. 
It does not have any effect on the production version.

### Screenshots

_No response_

### Additional information

_No response_

mustaphaturhan:
#243 is blocker for this migration, we should merge it first.

## 👨‍💻 Changes proposed

- Migrates project from react-scripts to vite with backward compatible options. We can change the settings and remove some plugins on the way.

Closes #171


nilsnolde:
Wow, vite is crazy fast, how nice:) It builds and runs as expected on my local (even freshly cloned) setup. 

Hm, I wonder why `valhalla1.openstreetmap.de` (note the 1, that's the HTTP backend) isn't working for you. For whatever reason we have an actual `.env` checked in, as long as you're not changing anything it should work out-of-the-box. The backend has CORS disabled.

nilsnolde (approved):

PR #246, opened 2025-09-18
Rewrite components with functional components

### LINKED ISSUE #68: Rewrite using functional components

mustaphaturhan:
Closes #68 

## 👨‍💻 Changes proposed

> While React class components are not going anywhere any time soon, functional components have become the new norm. A rewrite using functional components and hooks could make the application future proof and might offer a way to simplify the code along the way (especially the React <-> Leaflet interactions, which are all buried inside one big Map component).

### DIFF
--- a/.env
+++ b/.env
@@ -7,3 +7,4 @@ REACT_APP_CENTER_COORDS="52.51831,13.393707"
 # uncomment this variable to set boundaries on the leaflet map
 #                       southwest corner, northeast corner
 REACT_APP_MAX_BOUNDS="-90,-1e7,90,1e7"
+ESLINT_NO_DEV_ERRORS=true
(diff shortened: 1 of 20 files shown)


nilsnolde:
No worries @mustaphaturhan , I'm very amazed how quickly you're doing these things, it's hard to keep up with reviewing😄but we're very very happy things are finally improving! 

I'll have a glance later and run the app a bit, the ship it.

mustaphaturhan:
@nilsnolde all of the problems you mentioned should be fixed now, can you check them again please? 🙏🏻

nilsnolde:
> i am not sure if we are talking about the same thing but i didn't wanted to fix this yet because we have this bug in production too

Oops, didn't notice that yet, but probably I forgot about an old issue. I guess now it just became apparent visually with the initial bug in map.jsx. No need to fix this in the refactor, I agree. But if it's done already, then also good:)

I'll have a look later again, thanks!

nilsnolde (approved):
Looks great, thanks a lot!

PR #250, opened 2025-09-27
chore: added paths-ignore instead of paths

nilsnolde:
just a tiny thing annoying me. but it's good, I wanted to test the setup of the GHA to deploy a preview, which should work again :crossed_fingers:

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,9 +1,11 @@
 name: Tests and Linting
 on:
-  push:
-    branches: [main, master]
   pull_request:
     branches: [main, master]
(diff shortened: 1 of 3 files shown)


.github/workflows/playwright.yml
```diff
@@ -1,9 +1,11 @@
 name: Tests and Linting
 on:
-  push:
-    branches: [main, master]
```
nilsnolde, line 4: I'd protect master and only go over PRs. then tests don't need to run on master pushes

mustaphaturhan (approved):
thanks! looks fine to me. it's nice to see deploy working again.

PR #251, opened 2025-09-27
revert a bit of the changes..

nilsnolde:
oops, I changed too much, this one line was making sure the app dir is deleted again on the server.

### DIFF
--- a/.github/workflows/preview.yml
+++ b/.github/workflows/preview.yml
@@ -2,14 +2,12 @@ name: PR preview
 
 on:
   pull_request:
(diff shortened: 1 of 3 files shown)


.github/workflows/preview.yml
```diff
@@ -3,13 +3,12 @@ name: PR preview
 on:
   pull_request:
     types: [opened, reopened, synchronize, closed]
```
nilsnolde, line 5: haha is github broken? I mean, I used the same branch as before, but that was already deleted remotely. this is the actually recovered line, the rest is already in master AFAIK :)

PR #255, opened 2025-09-28
refactor: use kebap-case and make it a rule

mustaphaturhan:
## 🛠️ Fixes Issue

Not specific issue. Just make naming consistent everywhere.

## 👨‍💻 Changes proposed

- Enforce kebab-case for file names (and most folders) using eslint-plugin-check-file.
- Add small overrides so tests (*.spec|*.test.*) and type files (*.d.ts, vite-env.d.ts) not complain.
- Rename all mixed-case files to kebab-case

## 📄 Note to reviewers

Big diff look scary but mostly file moves. Real code change is tiny.


mustaphaturhan:
it looks like I forgot ignore some file formats from es linting, I updated the code to fix build failure @nilsnolde

nilsnolde (approved):

PR #259, opened 2025-10-01
feat: more e2e tests

tristancorbellari:
## 👨‍💻 Changes proposed

Added some more e2e tests, mostly to do with the left drawer and search functionality (which was a bit lacking in the tests suite), as well as some other smaller things (i.e. asserting visibility of waypoint markers on maps).

## 📄 Note to reviewers

Would be happy to implement more of these, but might just need to have a think about what the most important parts of the application would be to test, and how to avoid introducing a lot of redundancy.

### DIFF
--- a/e2e/map.spec.ts
+++ b/e2e/map.spec.ts
@@ -504,32 +504,27 @@ test.describe('Map interactions with left context menu', () => {
     await expect(page.getByTestId('dd-button')).toContainText(
       '13.393707, 52.517892'
     );
-    await expect(page.locator('.ui.compact.icon').first()).toBeVisible();
+    await expect(page.getByTestId('dd-copy-button')).toBeVisible();
 
     await expect(page.getByTestId('latlng-button')).toContainText(
       '52.517892, 13.393707'
     );
-    await expect(
-      page.locator('.mt1 > .ui.tiny > .ui.compact.icon').first()
-    ).toBeVisible();
+    await expect(page.getByTestId('latlng-copy-button')).toBeVisible();
 
     await expect(page.getByTestId('dms-button')).toContainText(
       '52° 31\' 4" N 13° 23\' 37" E'
     );
-    await expect(
-      page.locator('div:nth-child(3) > .ui.tiny > .ui.compact.icon')
-    ).toBeVisible();
+    await expect(page.getByTestId('dms-copy-button')).toBeVisible();
 
     await expect(
       page.getByRole('button', { name: 'Locate Point' })
     ).toBeVisible();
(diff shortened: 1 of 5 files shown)


e2e/map.spec.ts
```diff
@@ -373,25 +501,37 @@ test.describe('Map interactions with left context menu', () => {
...
-    ).toBeVisible();
-
-    await expect(
-      page.getByRole('button', { name: 'Valhalla Location JSON' })
-    ).toBeVisible();
-
     await expect(page.getByTestId('dd-button')).toContainText(
       '13.393707, 52.517892'
     );
+    await expect(page.locator('.ui.compact.icon').first()).toBeVisible();
```
mustaphaturhan, line 507: can we add data-testid or meaningful class names to this elements? `.ui.compact.icon` doesn't mean anything unfortunately.

e2e/map.spec.ts
```diff
@@ -373,25 +501,37 @@ test.describe('Map interactions with left context menu', () => {
...
 
     await expect(page.getByTestId('latlng-button')).toContainText(
       '52.517892, 13.393707'
     );
+    await expect(
+      page.locator('.mt1 > .ui.tiny > .ui.compact.icon').first()
```
mustaphaturhan, line 513: same here. this type of things can be change with improvements all the time like whenever we updated the margin of first container. when it happens, this test will fail even though it is only a ui change. it might be better if we name this buttons/icons better.

e2e/map.spec.ts
```diff
@@ -431,6 +571,23 @@ test.describe('Map interactions with left context menu', () => {
...
   });
+
+  test('should copy text to clipboard', async ({ page }) => {
+    await setupHeightMock(page);
+
+    await page.getByTestId('map').click({
+      button: 'left',
+    });
+
+    await page.locator('.ui.compact.icon').first().click();
```
mustaphaturhan, line 582: same here. and also, if we can't select a button or an icon with `page.getByRole`; that means our application is not accessible enough. this might be a good point to start to think about accessibility.

mustaphaturhan (changes requested):
thanks for adding more tests! this will definitely make the application more safer than ever. i have some notes though.

e2e/map.spec.ts
tristancorbellari: Good catch, already added a bunch of `data-testid`s throughout but missed a few. Just added those ones

tristancorbellari:
Tried to remove usage of `page.locator` as much as possible, except with some Leaflet map elements where I'm not sure if there's an alternative. I think it would be good practice to stick to `getByTestId`, `getByRole` and maybe `getByText` as much as possible

PR #264, opened 2025-11-04
Add switcher for carto and shortbread styles

### LINKED ISSUE #263: Implement layer picker and include OSM carto raster again

Not sure when, or even if, this has changed, but when tracing routes, the displayed map seems to have changed from the detailed OSM map to something very (too) basic, with one of my my biggest issues the fact that for motorways the road number no longer shows.

mustaphaturhan:
- Adds a new way to switch between carto and shortbread styles.

## 🛠️ Fixes Issue

Closes #263 

## 👨‍💻 Changes proposed

- Adds a new switch to settings panel:
image
- Adds `map_style` query parameter. `map_style=shortbread` will show new osm vector map. `map_style=carto` will show good, old raster map.

## 📷 Screenshots

image
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,6 +18,7 @@
         "d3": "^7.9.0",
         "date-fns": "^2.29.3",
         "json-format": "^1.0.1",
+        "lucide-react": "^0.552.0",
         "maplibre-gl": "^5.8.0",
         "prop-types": "^15.8.1",
         "ramda": "^0.28.0",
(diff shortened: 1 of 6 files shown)


mustaphaturhan:
@nilsnolde make sense, I updated the code:
image
image

nilsnolde (approved):
this is great, thanks for fixing it so quickly!

nice touch with the layer preview, I see they're their own map, it's a funny effect when one moves the map while the layer preview is open:)

prino:
When can we expect this to be generally available, the vector map is, excusez-le-mot, a total sh*t-show!

mustaphaturhan:
@prino this is already generally available. you can switch layers by using this button in the map.
image

or you can use: https://valhalla.openstreetmap.de/?style=carto

Issue #256, opened 2025-09-28
use better user interface and drag and drop library to unblock react 18 upgrade (maybe shadcn?)


mustaphaturhan:
### What kind of feature you are suggesting?

Currently we are using `semantic-ui-react`, `semantic-ui-css`, `@mui/material`, `tachyons` and some tools like `react-modern-drawer`, `@emotion/react` and `@emotion/styled`.

The most important part is semantic-ui-react because it is very outdated and doesn't support react 18. And to be honest, I don't think we need them.

Since we are -kind of- beginning of the project, we can use something like [shadcn](https://ui.shadcn.com/docs), a headless user interface library that helps you create your own design library. I am also willing to do this.

### Does it solve any problem?

- It will unblock react 18 upgrade.
- It will make it look better and more modern than it is now.

PR #268, opened 2025-11-26
Refactor everything

mustaphaturhan:
Related with #256.

## 🛠️ Fixes Issue

- Updates to react 18
  - Migrates from `react-beautiful-dnd` to `@hello-pangea/dnd` since `react-beautiful-dnd` is not supporting anymore. In the future, we will migrate to better dnd library but `@hello-pangea/dnd` is full compatible with `react-beautiful-dnd`
  - Changes `render` with `createRoot` since it is deprecated.
- Removes `semantic-ui-react` and starts to use `tailwindcss`
  - `semantic-ui-react` is very outdated library and we have better alternatives at the moment. Instead of it, we are starting to use our design system library that created with **shadcn**. So we will be able to freely change everything.
  - Packages related with material-ui also removed. There is no `emotion` now.
- Migrate away from `mapStateToProps` in favor of react hooks. Use `useSelector` and `useDispatch` instead of it. However, it is not finalized yet. We may completely deprecate it in future updates.
- Refactor the map component and make it more readable
- Updates UI from start to end.
- It won't wait reverse geocode to resolve the route if `wps` exist in the URL.
- New searching experience
- New settings experience
- Updates dependencies
- Renames the app name with `valhalla-web` and makes version 1.0.0
- Moves components under `components` folder for better, organized structure
- Removes old, create-react-app leftovers and migrates them to proper vite format. To make it happen, also updates the `.env` file.
- Updates package scripts for better organization and added husky pre-commit hook for linting.

**This pull request is still work in progress but I wanted to create it to get some opinions from people. So if you have some recommendations or objections, please let me know** 🙏🏻.

## Todos

- [x] Update the right, settings panel with new components
- [x] Fix race conditions
- [x] When user change the profile, send the request for both isochrones and routes. Otherwise we are showing "car" as a selected but "pedestrian" as a route.
- [x] When user change the isochrones tab and refresh the page, we are losing waypoint

## Follow-ups
- [ ] Currently, we are sending a request whenever small portion of settings changed. Like the change might be intentional or not. So instead of fetching as soon as user changed the value, we should give an option like "save" and we should send the request whenever user save it. It is better for handling huge loads.
- [ ] Start to use `tanstack/query` for request for better caching support. For example, reverse geocode responses aren't changing a lot. So we can cache the values.
- [ ] Add unit tests.

## 📄 Note to reviewers

Good luck. Sorry for this huge pull request :(

## 📷 Screenshots

image
image
image
image
image

### DIFF
--- a/src/actions/isochrones-actions.ts
+++ b/src/actions/isochrones-actions.ts
@@ -16,7 +16,7 @@ import {
 import { toast } from 'sonner';
 import { VALHALLA_OSM_URL, buildIsochronesRequest } from '@/utils/valhalla';
 
-import { showLoading, filterProfileSettings } from './common-actions';
+import { showLoading, filterProfileSettings, zoomTo } from './common-actions';
 import { calcArea } from '@/utils/geom';
 import type {
   ActiveWaypoint,
@@ -176,6 +176,7 @@ export const fetchReverseGeocodeIso =
   (lng: number, lat: number): ThunkResult =>
   (dispatch) => {
     dispatch(placeholderAddress(0, lng, lat));
+    dispatch(zoomTo([[lat, lng]]));
 
     dispatch({
       type: REQUEST_GEOCODE_RESULTS_ISO,
--- a/src/components/map/draw-control.tsx
+++ b/src/components/map/draw-control.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import { useRef, useMemo } from 'react';
 import { useControl } from 'react-map-gl/maplibre';
 import { MaplibreTerradrawControl } from '@watergis/maplibre-gl-terradraw';
 import '@watergis/maplibre-gl-terradraw/dist/maplibre-gl-terradraw.css';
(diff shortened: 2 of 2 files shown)


MoonE:
After deleting a location the `Directions` heading is still visible.
image

The location popup should disappear when right clicking in the map, instead it changes its position to and remains visible behind the new popup.
image

MoonE:
After selecting one waypoint both delete buttons become enabled:
image


There is a problem with the layout for the isochrone result. The Export button is not fully visible and there are some line breaks.
image

Previously it looked like this:
image

mustaphaturhan:
> The timestamp is no longer fully visible.

This is a problem because regions might have different formats for dates. For example, this is how it looks like in my region:
image

I created a workaround that makes select field width dynamic according to selected value but in the future, we might want to use a real date picker instead of relying browser's default.

I will create new issues for these follow-ups after we merge this PR. 

> An unnecessary request is sent when clicking into a numberfield and then clicking somewhere else without changing it.

This is fixed now.

mustaphaturhan:
@nilsnolde 

>  if a polygon is drawn it works fine, but if it's deleted the next request will still send the deleted polygon(s).

It looks like my logic for draw controller wasn't working for removing events. I updated the logic to make it run when user deleted something. This should be fixed now.

> at least on mobile I can't set a isochrone center on the map, only via geocoding. Didn't try on a computer, it's probably just a maplibre event missing?

I am not able to reproduce this one though. It works when you touch little bit long.

https://github.com/user-attachments/assets/9fe2b0ed-c69f-4621-ab1b-0826539a9d15

PR #271, opened 2025-12-06
Deprecate redux and introduce zustand. Use tanstack/query for async requests.

### LINKED ISSUE #257: deprecate redux and introduce zustand or redux/toolkit. and use tanstack/query for requests.

Currently, we are using old redux version which is not have "amazing" typescript support. We have to write every type manually which is not good at all. There are better options at the moment like zustand or redux/toolkit.

We can use:
- @reduxjs/toolkit which already have rtk query. We can build queries by using @reduxjs/toolkit/query/react which creates automatic hooks. Or,
- We can introduce zustand for global state management which is more easier than redux since it has lesser boilerplate and tanstack/query for request management which has amazing cache support.
- We can completely get rid of global state management but I can't say we don't need it because we want to control map from other components as well and tanstack/query for request management which has amazing cache support.

- Better cache management
- More modern structure for a single page application

### Screenshots

_No response_

### Additional information

_No response_

mustaphaturhan:
Closes #257

## 🛠️ Fixes Issue

- Redux comes with huge boilerplate. However, there is more easier way to handle things. **Zustand**. More easy to use and modern.
- Moves async functions to tanstack/query.

#### Why `useQuery with refetch`; instead of `useMutation`.

Caching. Even though we are not using the important keys as a queryKey at the moment; it will changed in the future. Let's say user search for "Berlin", and then search something else, and then searched "Berlin" again. We should be able to show the results instantly. Without going to server.

Nominatim, valhalla and other APIs have very limited source. We should be careful while using them.

Just note: Caching is not implemented in this pr. It will be separate.


nilsnolde:
wonder why gha is asking for a valid email address now when deploying the app.. (nothing to do with this PR of course, I will solve that separately)

mustaphaturhan:
the problem is not related with deployment though. deployment is here: https://valhalla-app-tests.gis-ops.com/271. probably something is wrong with commenting.

PR #273, opened 2025-12-07
Update marker anchor and styling in map component

### LINKED ISSUE #272: Location arrows completely overlay points
### Description

Since now (i.e. two, three days), the web interface has been changed, and not for the better. My biggest issue id the fact that the location markers now completely cover the locations, and even at the largest magnification, the pointy arrow if the marker in not on the exact location, obtained via OSM, the "Internal" location, of the coordinates. 

Please restore the original way of overlaying the markers!

### Screenshots

Image

mustaphaturhan:
Closes #272 

## 🛠️ Fixes Issue

- Adds anchor to marker.


nilsnolde (approved):
looks much better, thanks!

PR #277, opened 2025-12-26
Add new basemap

### LINKED ISSUE #127: [FEATURE] New basemap(s)

Let's get Stadia's Alidade base map as the default (and keep OSM for the time being).

OSM base map is not good for routing.

### Screenshots

_No response_

### Additional information

https://stadiamaps.com/themes/

Need to generate a API key. This is best done via Github Actions so we can encode the API key there. Needs a Github Action first though:)

mustaphaturhan:
Closes #127 

## 🛠️ Fixes Issue

- Adds new alidade smooth style
- Improves the logic of adding new style and makes the constant only source of truth
- Imports json files from direct urls instead of importing them directly

## 📷 Screenshots

image

### DIFF
--- a/src/components/map/constants.ts
+++ b/src/components/map/constants.ts
@@ -23,13 +23,15 @@ export const routeObjects = {
 export const MAP_STYLE_STORAGE_KEY = 'selectedMapStyle';
 export const CUSTOM_STYLE_STORAGE_KEY = 'customMapStyle';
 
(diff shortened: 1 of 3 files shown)


nilsnolde:
I created an account for the Valhalla project an pointed it to valhalla.openstreetmap.de. I'll set up a password manager for the project too so we can easier share logins.

PR #285, opened 2026-01-03
Fallback profile parameter to default profile

### LINKED ISSUE #284: URL to `/isochrones` & `/directions` is erroring
### Description

when calling https://valhalla.openstreetmap.de/isochrones or /directions, we get 

`Cannot read properties of undefined (reading 'boolean')`

I'd have assumed this succeeds and chooses the right tab.

### Screenshots

_No response_

mustaphaturhan:
## 🛠️ Fixes Issue

Closes #284

## 👨‍💻 Changes proposed

- We are using url parameter "profile" as our source of truth while settings the profile. It should be always defined. This PR adds fallback so when there is no profile in the URL, it will use 'bicycle'.

### DIFF
--- a/e2e/homepage.spec.ts
+++ b/e2e/homepage.spec.ts
@@ -9,6 +9,39 @@ test('has title', async ({ page }) => {
   await expect(page).toHaveTitle(/Valhalla FOSSGIS/);
 });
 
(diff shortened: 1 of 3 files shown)


nilsnolde (approved):
thanks for the quick action:)

PR #292, opened 2026-01-06
Add optimize route button

### LINKED ISSUE #287: Give "optimize route" option to user

Currently, we create routes strictly in the order added. However, `valhalla` provides an `optimized_route` service that automatically reorders waypoints for the most efficient path.

So why not use it?

We can show an 'optimize route' button when there are more than 2 waypoints to trigger this endpoint.

It demonstrates more features of valhalla

### Screenshots

_No response_

### Additional information

_No response_

mustaphaturhan:
## 🛠️ Fixes Issue

Closes #287 

## 👨‍💻 Changes proposed

- Adds new optimize route button -which is disabled by default. If there is more than 4 route, then the button will be enabled.
- When user click the button, it will send a request to `optimized_route` endpoint. It will update the waypoints and route according to result.

## 📷 Screenshots

image
image

### DIFF
--- a/src/components/directions/directions.spec.tsx
+++ b/src/components/directions/directions.spec.tsx
@@ -68,6 +68,13 @@ vi.mock('@/hooks/use-directions-queries', () => ({
   })),
 }));
 
+vi.mock('@/hooks/use-optimized-route-query', () => ({
+  useOptimizedRouteQuery: vi.fn(() => ({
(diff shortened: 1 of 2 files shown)


mustaphaturhan:
@chrstnbwnkl I believe there are only two profiles enabled for `optimized_route` but I can't remember from my head directly. maybe @nilsnolde might know better?

PR #295, opened 2026-01-07
fix: add github comment to 'about this service' link in attribution

nilsnolde:
## 👨‍💻 Changes proposed

for the shortbread tiles we were linking to OSRM's "About" page. I changed that to this github comment where we introduce the service: https://github.com/valhalla/valhalla/discussions/3373

### DIFF
--- a/public/styles/carto.json
+++ b/public/styles/carto.json
@@ -7,7 +7,7 @@
       "tileSize": 256,
       "minzoom": 0,
       "maxzoom": 19,
-      "attribution": "&copy; <a href=\"https://www.openstreetmap.org/copyright\">OpenStreetMap</a> contributors"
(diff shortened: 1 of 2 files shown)


nilsnolde:
Do you mean higher API limits? 

Or maybe you don't know that the API is fully openly accessible?

mustaphaturhan:
> Or maybe you don't know that the API is fully openly accessible?

i misunderstood what I read 🤦🏻 i knew it was open to everyone and i used it in an app I was working on... but let's call it morning brain fog :)

> we rate limit to 1 call/user/sec and 100 calls/sec total. Service limits are a bit stricter than the defaults as well. That's about all one needs to know.

 this is quite good for most applications.

PR #299, opened 2026-01-08
test: add previews deployments for forks

mustaphaturhan:
When a new pull request opened from fork repositories, we are not able to deploy it since our secret key only exist in `valhalla/web-app` repository. This is just a test to make it work.


mustaphaturhan:
@nilsnolde can we merge and test this? deploy pipeline is not working since it is not on the master.

nilsnolde (approved):

PR #302, opened 2026-01-09
test new pipelines from remote

mustaphaturhan:
let's see if we can deploy a preview from remote


mustaphaturhan:
awesome. but refresh is not working again because of htaccess or something else. i will check it.

PR #306, opened 2026-01-09
Introduce more workflow updates

mustaphaturhan:
- Upgraded actions/checkout from v2 → v4
- Upgraded actions/setup-node from v2 → v4
- Instead of `npm install`, use `npm ci` in the deploy pipeline
- Add npm caching in deployment and preview builds
- Add concurrency control and set `cancel-in-progress`: false. It will prevent race conditions when multiple commits are pushed in very short time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,8 +1,14 @@
 {
   "name": "valhalla-web",
   "version": "1.0.0",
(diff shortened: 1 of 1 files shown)


nilsnolde (approved):
thanks, that's great!

PR #310, opened 2026-01-14
feat: implement runtime base url configuration

### LINKED ISSUE #308: Add option for base URL

Let the app user decide the base URL of the Valhalla endpoints, so we can easily switch to localhost for debugging. That goes for routing, isochrones (and likely soon for MVT: #307). 

### Screenshots

_No response_

### Additional information

We might come up with more settings in the future (though I have nothing specific in mind at the moment). Or move some of the settings in the right panel, like whether to geocode or not, or language settings.

mustaphaturhan:
## 🛠️ Fixes Issue

Closes #308 

## 👨‍💻 Changes proposed

- Makes sections in the settings panel collapsible.
- Adds new server settings to change base url in the runtime.
  - We are sending a request to `<input-url>/status` endpoint after user changed the URL. It throws an error if the URL is not in a URL format, or it is not reachable, or doesn't respond in 5 seconds or doesn't return correct response. If we get a response, we are checking route and isochrone in `available_actions`. If they are not presented in the result, we also doesn't let the change of base url 
- We are saving the changed base url to local storage. If user refresh the page, they will continue to see new base url they set.
- There is also a "reset base url" button in the panel. If user clicks it, the base url will return back to the value set in the `VITE_VALHALLA_URL` environment variable.

## 📄 Note to reviewers

I haven't given much thought to security but since everything happens in the user's computer, I don't think there will be a problem.

## 📷 Screenshots

image
image
image
image

### DIFF
--- a/src/components/settings-panel/server-settings.tsx
+++ b/src/components/settings-panel/server-settings.tsx
@@ -1,4 +1,5 @@
 import { useState } from 'react';
+import { useMutation } from '@tanstack/react-query';
 import { Button } from '@/components/ui/button';
 import { Input } from '@/components/ui/input';
 import {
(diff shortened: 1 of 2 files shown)


nilsnolde (approved):
again thanks for the super quick implementation:)

> Makes sections in the settings panel collapsible.

nice touch!

Issue #307, opened 2026-01-12
dedicated `/tile` UX to (de-)select layers & attribute popups on feature left-click


nilsnolde:
### What kind of feature you are suggesting?

For https://valhalla1.openstreetmap.de/tile, it'd be great to have a dedicated UX for:
- (de-)selecting single layers: we now have `edges` & `layers`, but I also plan on adding a few more layers, like `shortcuts` & `graph` (to show valhalla routing tile extents, although that wouldn't _necessarily_ need an endpoint, graph tiles have static size, but it's still nicer IMO)
- popup on left-click on a MVT feature where we can see all attribute values of an edge or node
- others?

It could be another map control which just shows a menu to (de-)select all layers (all on by default). 

### Does it solve any problem?

We now have custom styles since a while and that's already working of course to add MVT from Valhalla's `/tile` endpoint. However, that makes it a bit cumbersome for using the app easily for debugging routing engine issues., e.g. (de)selecting single layers, offering click events/popups.

WDYT @mustaphaturhan ?

### Additional information

Since a while we have [server-side MVT caching](https://github.com/valhalla/valhalla/pull/5762) enabled which was necessary for us to allow public requests to https://valhalla1.openstreetmap.de/tile. I'll enable it on the server now-ish. After merging https://github.com/valhalla/valhalla/pull/5809 it'll be even better for the service.

Also note, that we soon have a [filtering option for included feature attributes](https://github.com/valhalla/valhalla/pull/5806), which will default to a very small subset of available attributes. For the debugging functionality in the app, it'll be necessary to use `"verbose": true` in the request, e.g.

```
/tile?json=%7B%0A%09%22tile%22%3A%20%7B%0A%09%09%22z%22%3A%20{z}%2C%0A%09%09%22x%22%3A%20{x}%2C%0A%09%09%22y%22%3A%20{y}%0A%09%7D%2C%0A%09%22verbose%22%3A%20false%2C%0A%09%22tile_options%22%3A%20%7B%0A%09%09%22return_shortcuts%22%3A%20false%0A%09%7D%0A%7D
```

the (MVT pattern compatible) URL encoded version of

```json
{
	"tile": {
		"z": {z},
		"x": {x},
		"y": {y}
	},
	"verbose": true
}
```

I'd also like to add some labeling to the layers, but that can be entirely defined in the style.json.

PR #311, opened 2026-01-15
Add tiles editor to hide/show layers

mustaphaturhan:
## 🛠️ Fixes Issue

Related with #307 but it is not completely fix the problem. It adds new tiles tab to the homepage, and gives capability of making layers visible/invisible.

## 👨‍💻 Changes proposed

- Adds new tiles tab
- Lists the layers of `style.json`
- Adds search input for layers

## 📷 Screenshots

image


nilsnolde (approved):
in any case, this PR is good as it is:)

PR #312, opened 2026-01-18
Give an option to append valhalla tiles to `style.json` file

mustaphaturhan:
## 🛠️ Fixes Issue

Related with #307.

## 👨‍💻 Changes proposed

- Adds new "Append Valhalla layers" switch to the top of tiles panel
- When it is selected, it also lists the prepared layers in the layer list, so user can select/de-select visibility of the layer.

## 📄 Note to reviewers

- In localhost, we are getting a lot of `CORS` errors. Probably it will happen in our demo too.
- I believe we set the max zoom level as 14 in the back-end. When we zoom a lot, the back-end throws `error 400` with a response: `{"error_code":175,"error":"Exceeded max zoom level of: 14","status_code":400,"status":"Bad Request"}`
- When we add new layers to `tile` server, we need to update `src/components/tiles/valhalla-layers.ts` file manually.

## 📷 Screenshots

image
image

### DIFF
--- a/src/components/tiles/valhalla-layers.spec.ts
+++ b/src/components/tiles/valhalla-layers.spec.ts
@@ -154,7 +154,7 @@ describe('valhalla-layers', () => {
       const spec = getValhallaSourceSpec() as VectorSourceSpecification;
 
       expect(spec.minzoom).toBe(7);
-      expect(spec.maxzoom).toBe(22);
+      expect(spec.maxzoom).toBe(14);
     });
 
     it('should use xyz scheme', () => {


mustaphaturhan:
@nilsnolde my browser says `No 'Access-Control-Allow-Origin' header is present on the requested resource.` and my humble opinion is, this happens because you are doing `add_header "Access-Control-Allow-Origin"  *;` only if `$request_method = OPTIONS`. maybe:

```nginx
location ~ ^/(route|optimized_route|status|isochrone|trace_route|trace_attributes|sources_to_targets|expansion|height|locate|tile) {
    # Add CORS headers to ALL responses
    add_header "Access-Control-Allow-Origin" * always;
    add_header "Access-Control-Allow-Methods" "GET, POST, OPTIONS, HEAD" always;
    add_header "Access-Control-Allow-Headers" "Authorization, Origin, X-Requested-With, Content-Type" always;

    # Preflighted requests
    if ($request_method = OPTIONS) {
        add_header "Access-Control-Allow-Origin" * always;
        add_header "Access-Control-Max-Age" 86400;
        add_header "Access-Control-Allow-Methods" "GET, POST, OPTIONS, HEAD";
        add_header "Access-Control-Allow-Headers" "Authorization, Origin, X-Requested-With, Content-Type";
        return 204;
    }

    limit_req zone=valhalla_per_ip_rate burst=10;
    limit_req zone=valhalla_global_limit;
    limit_req_status 429;
    proxy_pass http://valhalla2.openstreetmap.de;
    proxy_set_header Host valhalla2.openstreetmap.de;
}
```
would fix the problem, but I am not sure since my devops capabilities are low :( this is a straight answer from opus 4.5. 

> what we really want is that maplibre requests level 14 tiles for anything beyond level 14

I see now... so the thing is, I was porting our default, recommended [style.json file](https://github.com/valhalla/valhalla/blob/master/docs/docs/api/tile/default_style.json) directly and we are defining our styles like this:
```json
"sources": {
    "valhalla": {
      "type": "vector",
      "tiles": [
        "http://localhost:8002/tile?json=%7B%22tile%22%3A%7B%22z%22%3A{z}%2C%22x%22%3A{x}%2C%22y%22%3A{y}%7D%7D"
      ],
      "minzoom": 7,
      "maxzoom": 22, // this is problematic because we are only supporting maxzoom level 14.
      "scheme": "xyz"
    }
},
```

I updated `maxzoom` with `14`, now we are not sending requests for zoom levels > 14 at all and it saves the day.

PR #321, opened 2026-01-29
fix: /tile request should be with verbose = true

### LINKED ISSUE #320: add `"verbose": true` to the /tile request
### Description

with https://github.com/valhalla/valhalla/pull/5806, we added options to filter tiles and changed the default to only return style-relevant attributes for `edges` & `nodes`, so that people can opt into larger tile sizes if they need to. 

for our web map do we want to, all we need to do at this point is add that request parameter and it'll stay the exact same. already as a warning: soon we merge https://github.com/valhalla/valhalla/pull/5849 which adds a new layer `shortcuts`.

### Screenshots

_No response_

### Additional information

_No response_

nilsnolde:
## 🛠️ Fixes Issue

fixes #320 

currently https://valhalla.openstreetmap.de/tiles?profile=bicycle&style=carto shows only 2 edge attributes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,6 @@
 
 <img width="2253" height="1303" alt="image" src="https://github.com/user-attachments/assets/00f43ea8-51df-4319-ad31-4feadf0024c1" />
 
-
 This is the ReactJS demo web app running on https://valhalla.openstreetmap.de. It provides routing and isochrones with a magnitude of options and makes requests to [Valhalla](https://github.com/valhalla/valhalla), an open source routing engine and accompanying libraries for use with OpenStreetMap data.
 
 ## Commands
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -27,6 +26,12 @@ This project includes end-to-end tests using [Playwright](https://playwright.dev
 
 ### Running Tests
 
+First install the `playwright` browsers:
+
+```bash
+npx playwright install
```
nilsnolde: `--with-deps` was causing some error. arch linux doesn't seem to be natively supported for playwright, it's installing the fallback ubuntu browsers. for some reason it only worked without that argument

nilsnolde:
actually never mind;)

Issue #328, opened 2026-01-31
should we filter boolean MVT attributes to the ones that matter?


nilsnolde:
### What would you like to share?

I already enjoy the popups for Valhalla dev a lot! but it is a bit overwhelming bcs there's so many attributes..

I wonder if we shouldn't filter the `false` ones which are expected to be false, e.g. `toll`, `deadend` etc. most of the booleans are important to understand when they're `true`, hardly any `false` is valuable information.

wdyt @chrstnbwnkl ?

PR #329, opened 2026-01-31
filter falsey bools & live traffic

nilsnolde:
trying out #328 for feedback.

wdyt?

### DIFF
--- a/src/components/settings-panel/settings-options.ts
+++ b/src/components/settings-panel/settings-options.ts
@@ -1,4 +1,4 @@
-import type { BicycleType } from '@/components/types';
+import type { BicycleType, PedestrianType } from '@/components/types';
 
 interface NumericSetting {
   name: string;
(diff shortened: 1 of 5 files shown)


src/components/map/parts/tiles-constants.ts
```diff
@@ -140,3 +140,59 @@ export const bitmaskPropertyMappings: Record<string, Record<number, string>> = {
...
+  'named_intersection',
+];
+
+export function shouldHideProperty(key: string, value: unknown): boolean {
+  if (typeof value === 'boolean' && !value && hideWhenFalseKeys.includes(key)) {
+    return true;
+  } else if (
+    typeof value == 'number' &&
+    key.includes('speed') &&
+    (value == 0 || value == 254)
```
chrstnbwnkl: `value == 254` would also hide e.g. `breakpoint2 == 254` which could be valid right?

chrstnbwnkl: since it has the key `live_speed:<dir>:breakpoint2`

PR #330, opened 2026-01-31
feat: add more costing options

nilsnolde:
## 👨‍💻 Changes proposed

added a few costing options which are not yet implemented:
- `elevator_penalty`
- `type` for pedestrian: foot, wheelchair, blind

also moved a few more existing costing options to the pedestrian profile.


src/components/types.ts
```diff
@@ -180,27 +184,6 @@ export interface Center {
   addressindex: number;
 }
 
-export interface Costing {
```
nilsnolde: these were unused

PR #337, opened 2026-02-28
feat: add isochrone color palettes and opacity control

s4chinjha:


nilsnolde:
why did actually close here @s4chinjha ? I didn't look at the code but I did try UX and that looked pretty good IMO.

Issue #333, opened 2026-02-23
add `shortcuts` layer


nilsnolde:
since https://github.com/valhalla/valhalla/pull/5849 we removed shortcuts from the `edges` layer and instead created the `shortcuts` layer. the [default_style.json](https://github.com/valhalla/valhalla/blob/master/docs/docs/api/tile/default_style.json) has also been updated, cloning the style of the `edges` layer.

PR #338, opened 2026-02-28
feat: add shortcuts layer support

s4chinjha:
This Pr adds the **shortcuts** layer as mentioned in the issue #333 

**Addition**

Since shortcuts was removed from the edges layer and the shortcuts layer was created we need to add it here in the web-app.

**Changes added**

I have added the support for the **shortcuts** layer and have reused the existing edges styling, also included the shortcuts in `VALHALLA_LAYERS`.

I have used the additions added in the updated [default_style.json](https://github.com/valhalla/valhalla/blob/master/docs/docs/api/tile/default_style.json) as reference.

**Result**

Shortcuts is now visible on the map

I have added the images to show it's implementation

**Before**

Screenshot From 2026-02-28 19-18-14  

**After** (shortcuts has been  added)

Screenshot From 2026-02-28 19-18-25

### DIFF
--- a/src/components/map/index.tsx
+++ b/src/components/map/index.tsx
@@ -87,7 +87,6 @@ export const MapComponent = () => {
   const directionsPanelOpen = useCommonStore(
     (state) => state.directionsPanelOpen
   );
-  const settingsPanelOpen = useCommonStore((state) => state.settingsPanelOpen);
   const updateSettings = useCommonStore((state) => state.updateSettings);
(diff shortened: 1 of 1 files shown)


nilsnolde (changes requested):
LGTM. the only small thing we need to change is that the shortcut pop-up shows "Node" in the header:

Image

mustaphaturhan:
This looks fine to me, it is good to go after the title is fixed.

Issue #191, opened 2023-07-09
[Isochrones] Use a color scale that is colorblind friendly


wcedmisten:
### What kind of feature you are suggesting?
Currently the [isochrones demo](https://valhalla.openstreetmap.de/isochrones) uses a red-green color scale for the contours.

Unfortunately these colors are not very accessible for people who are colorblind, which affects around [8% of the male population](https://www.nei.nih.gov/learn-about-eye-health/eye-conditions-and-diseases/color-blindness).

Since the colors convey meaning (green = near, red = far), it would be great to use a color scheme that is more accessible for colorblind people.

### Does it solve any problem?
Better understanding of the isochrones for colorblind people.

### Proposed Solution
Personally I've used the [Viridis color scale](https://cran.r-project.org/web/packages/viridis/vignettes/intro-to-viridis.html) in [my own projects for isochrones](https://wcedmisten.github.io/nextjs-protomap-demo/isochrone), but there are other color scales that are colorblind friendly.

### Screenshots

![image](https://github.com/gis-ops/valhalla-app/assets/4565200/004227c1-90ff-4631-aa97-639daf2fa22c)
![image](https://github.com/gis-ops/valhalla-app/assets/4565200/59922539-622f-4732-b066-4b7078bda8a2)

### Additional information

https://blog.bioturing.com/2018/09/24/heatmap-color-scale/

chrstnbwnkl:
Sounds like a meaningful enhancement to me. In your own project, you're using a grayscale base map though, whereas we're using the far more colorful standard Mapnik style. Would you mind giving it a shot to see how this would look like in combination?

jakern:
I made an attempt at this because the opacity of many isochrones was bothering me. Branch is [here](https://github.com/jakern/valhalla-app/tree/viridis-iso-color) 

Example of original where many isochrones overlapping would make map unreadable.
![Image](https://github.com/user-attachments/assets/7466d15e-c53e-4431-ab5b-631441859a40)

New Gradiant example
![Image](https://github.com/user-attachments/assets/3a5bdd13-1453-42af-b30c-c245d3144974)

Wider angle to try to see if the is enough color variability to differentiate the colors.
![Image](https://github.com/user-attachments/assets/f41e70c9-1845-42a2-bb67-38e8d58a6356)

Example with fewer contours that highlights the problem of differentiation.
![Image](https://github.com/user-attachments/assets/cfa1d8c6-d568-4e0c-a943-589de3a62c99)

PR #340, opened 2026-03-01
added isochrone color palettes and opacity control

### LINKED ISSUE #283: Make isochrone color palette user definable
### Description

The default isochrone colors are way too opaque, also they're not very accessible, see https://github.com/valhalla/web-app/issues/191.

It'd be great if we could add an option for the user to choose from a few predefined color scales. I wouldn't go as far as complete color control with custom palettes, IMO it'd be enough to add viridis as @wcedmisten suggested, plus smth like the current one.

Another helpful setting would be opacity.

In terms of UX, I guess the best would be to add a control to the isochrone summary? 

Image

Or maybe more as a global setting, on the right side with the other map controls? We don't support multiple isochrones yet, but I could imagine we will at some point, color palette shouldn't be a per-isochrone setting IMO.

Or into the isochrone input parameters? With the non-intuitive behavior (for me at least) that color changes still work after the isochrone has been drawn.

Any other suggestions?

### Screenshots

_No response_

### Additional information

_No response_

s4chinjha:
This PR Closes #283 and adds an Isochrone Settings Section to the isochrone summary panel.

## 🛠️ Fixes Issue

As mentioned in issue #283, the default isochrone colors are way too opaque and are not very accessible (see also #191). There is also no way of changing opacity right now, which makes it harder to read the map underneath when multiple contours overlap.

## 👨‍💻 Changes proposed

**New file -  `src/utils/isochrone-palettes.ts`**
Added a utility that defines the available color palettes and handles color interpolation across any number of contours.

**Color Palette selector**
Added an option for the user to choose from a few predefined color scales - Default, Viridis (colorblind-friendly), Plasma, and Blues. 

The Default option keeps the existing API-provided colors so nothing changes for users who don't touch the setting.

**Opacity slider**
Added a slider (0–1) to control fill transparency. Useful when contours overlap or when the map details underneath need to stay readable.

**Both controls update the map live** - no need to redraw the isochrone after changing palette or opacity.

**Palette is a global setting**, not per-isochrone, so it will continue to work correctly when multiple isochrones are supported in the future.

**Fixed rendering order** so smaller contour rings are always visible on top of larger ones instead of being hidden underneath.

**Updated tests** to include the new store fields so all existing tests continue to pass.

I have used AI in assisting me

### DIFF
--- a/src/components/heightgraph.tsx
+++ b/src/components/heightgraph.tsx
@@ -239,7 +239,7 @@ const HeightGraph: React.FC<HeightGraphProps> = ({
     return () => {
       tooltip.remove();
     };
-  }, [data, dimensions, onHighlight]);
+  }, [data, dimensions, onHighlight, isExpanded]);
 
   const [prevProps, setPrevProps] = useState({ width, height });
   if (prevProps.width !== width || prevProps.height !== height) {


src/utils/isochrone-palettes.ts
```diff
@@ -0,0 +1,74 @@
...
+//Fallback fill color when the Valhalla API provides no color for a feature.
+export const DEFAULT_FILL = '#6200ea';
+
+//Default opacity for isochrone fill polygons.
+export const DEFAULT_OPACITY = 0.4;
+
+export interface IsochronePalette {
+  id: PaletteId;
+  label: string;
+  //Ordered from smallest to largest contour. null = use API-provided colors.
```
nilsnolde: I'm surprised `prettier` or whatever formatter allows this comment style. pls insert a space in between // and comment

```suggestion
  // Ordered from smallest to largest contour. null = use API-provided colors.
```

src/utils/isochrone-palettes.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+describe('getPaletteColor', () => {
+  it('returns DEFAULT_FILL for an empty palette', () => {
+    expect(getPaletteColor([], 0)).toBe(DEFAULT_FILL);
+  });
+
+  it('returns the only color when the palette has one entry', () => {
+    expect(getPaletteColor(['#aabbcc'], 0.5)).toBe('#aabbcc');
+  });
+
+  // t=1 is the largest contour, should map to the last color
```
mustaphaturhan: I mean the `it('description')` block is for comments you writing here. instead of comment here, you can write it in the description.

nilsnolde:
> I wanted to ask what workflow you would prefer in this situation. If I make several commits addressing review comments, should I keep them as separate commits instead of squashing them (to avoid force-pushing), or is squashing them into one commit acceptable?

@s4chinjha I think [my comment](https://github.com/valhalla/web-app/pull/336#issuecomment-3979814446) should explain that fairly ok. basically, _try to_ be clean: 1 commit per post-review fix (e.g. 4 change requests -> 4 post-review commits with a good commit message). it's not so bad if you break that, that's just for me the most ideal situation. but really, zero force pushes after initial review, never if you can.

s4chinjha:
@mustaphaturhan i could not see your suggestion before, the one you added 2 days ago.Most of the comments were for me i forgot to remove them really sorry.

src/stores/isochrones-store.ts
```diff
@@ -103,6 +113,17 @@ export const useIsochronesStore = create<IsochroneStore>()(
           'updateSettings'
         ),
 
+      updateVisualization: ({ colorPalette, opacity }) =>
+        set(
+          (state) => {
+            if (colorPalette !== undefined) state.colorPalette = colorPalette;
+            if (opacity !== undefined)
+              state.opacity = Math.min(1, Math.max(0, opacity));
```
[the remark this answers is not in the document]
s4chinjha: I can follow this up in a issue I was going to create for some questions/issues I wanted to ask.Then, from there I can work on this.

nilsnolde:
> I’ve fixed this by changing how colors are picked, each contour now gets a color based on its value relative to the largest contour, so the same contour always gets the same color even if the no. of contours changes.

hm, that's not true though, colors change depending on the no. of contours. I actually had in mind what you're writing here, so that all isochrones have a fixed color, e.g. largest one: max(color), 2nd largest one: max(color) - 1, 3rd largest one: max(color) - 2 etc. while I think that's better UX when changing intervals, the current PR implementation is more future proof, we should only need min & max of the palette.

PR #336, opened 2026-02-28
Fix: prevent map from re-zooming on UI state changes

s4chinjha:
This PR Fixes Issue #326

**Problem faced** 
The map was zooming back to its original position of when the route was just created 
1.if we open or close the directions panel it does the same thing. 
2. We scroll over to some other section in map even after Resetting the route.
The zoom was triggered both by state changes and route creation, so it kept running even when nothing actually changed.

**Changes made** 
This Pr fixes the issue where the map whenever opening or closing the direction panel  was zooming back  to the same state when the route was created and uses the latest panel state.

**Result**
Map zooms only once when a new route is created.

**Note** 
I would like to work on this further and develop a button that would help recenter to the route if the user wants too.

**Video of**

the problem faced

https://github.com/user-attachments/assets/94b26331-822e-48de-b5d5-9946a56c5cb7

after implementing the fix

https://github.com/user-attachments/assets/bfde6384-06a0-4f2b-a188-29d0f4ddc55a

I have used AI in assisting me.

Issue #326, opened 2026-01-31
Map zooms too agressively to route extents


nilsnolde:
### Description

I'd love to show some small video and I'm again astonished how this really simple task seems to be impossible on Arch Linux.. I want UI, no cli, no AUR, and _not_ OBS, aka impossible. Anyways, I'll try:

- when we compute a route, we zoom to it -> great
- if we zoom around after that and open/close either the left or right side bar, we also zoom back to it -> not so good
- even when we clear waypoints, it will keep zooming to the old route when open/close left/right sidebar

IMO ideally we'd only zoom to the route once it's created and disable that mechanism after. WDYT @mustaphaturhan ?

nilsnolde:
done in #336

PR #342, opened 2026-03-02
added the recenter-route feature control file

### LINKED ISSUE #339: Add a button that would recenter map to current route
**Description**

After the recent change to stop the map from automatically zooming to the route (ref : #336 ), users can pan or zoom away from the route and there is no easy way to return to it.

It would be useful to add a simple control to recenter the map to the current route when needed.

s4chinjha:
This Pr Closes #339 and adds a Recenter button to the map.

## 🛠️ Fixes Issue

As mentioned in issue #339, after Pr #336 disabled automatic map zooming, there was no way to return to the route once you pan or zoom away from it. So i wanted to add a simple control to recenter the map to the current route.

## 👨‍💻 Changes proposed
  - Added the recenter-route button to the map as propose in the issue
  - New button in the top-right MapLibre control group below Map Styles
  - Used a crosshair icon with the tooltip "Recenter to route"
  - Clicking the button fits the entire route back into view.
  - Button is hidden when no route is present and reappears automatically when a route is loaded.

## 📄 Note to reviewers

- I have reused the bounds calculation, padding logic, and fitBounds call from the zoom effect in `src/components/map/index.tsx`.
- The button was not hiding after resetting waypoints (same root cause as issue #326) which I fixed by checking successful from directionsStore instead of coordinates, since successful is set to false when clearRoutes() is called.

## 📷 Screenshots

A img just to show how it looks
image

I have used AI in assisting me

### DIFF
--- a/src/components/tiles/custom-layer-editor.spec.tsx
+++ b/src/components/tiles/custom-layer-editor.spec.tsx
@@ -1,0 +1,428 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import {
+  render,
+  screen,
+  waitFor,
+  fireEvent,
+  within,
+} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import type { LayerSpecification } from 'maplibre-gl';
+import { CustomLayerEditor } from './custom-layer-editor';
+
+const createMockMap = () => {
+  const layers: Record<string, unknown> = {};
+  return {
+    getLayer: vi.fn((id: string) => layers[id]),
+    addLayer: vi.fn((layer: { id: string }) => {
+      layers[layer.id] = layer;
+    }),
+    on: vi.fn((event: string, handler: () => void) => {
+      if (event === 'styledata') handler();
+    }),
+    off: vi.fn(),
+    getSource: vi.fn(() => ({ type: 'vector' })),
+    _layers: layers,
+  };
+};
+
+let mockMap = createMockMap();
+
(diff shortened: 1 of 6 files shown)


src/components/directions/summary.tsx
```diff
@@ -5,37 +5,93 @@ import type { Summary as SummaryType } from '@/components/types';
...
+          [firstCoord[1], firstCoord[0]],
+          [firstCoord[1], firstCoord[0]],
+        ]
+      );
+
+    const state = useCommonStore.getState();
+    const dpOpen = state.directionsPanelOpen;
+    const spOpen = state.settingsPanelOpen;
+
+    const paddingTopLeft = [screen.width < 550 ? 50 : dpOpen ? 420 : 50, 50];
```
mustaphaturhan: You can see that this `summary` component only being rendered inside directions panel. if there is no panel, then there won't be this button.

the calculations of `when panel is visible` and `not visible` is unnecessary.

src/components/directions/summary.tsx
```diff
@@ -5,37 +5,93 @@ import type { Summary as SummaryType } from '@/components/types';
...
+    const paddingBottomRight = [
+      screen.width < 550 ? 50 : spOpen ? 420 : 50,
+      50,
+    ];
+
+    mainMap.fitBounds(bounds, {
+      padding: {
+        top: paddingTopLeft[1] as number,
+        bottom: paddingBottomRight[1] as number,
```
mustaphaturhan: this is always 50.

PR #346, opened 2026-03-05
feat: add toast message for warnings array

### LINKED ISSUE #45: Add warnings from Valhalla
Since we implemented a `warnings` array for user feedback, we should find a way to display them. E.g. one of the most important information it returns is when we clamp the value of an input costing option to a default value when it's out of bounds, e.g. `top_speed` can only be 10 - 252 km/h and if it exceeds that range we clamp it to 140. That's important info for the user.

lourduradjou:
## 🛠️ Fixes Issue
Closes #45 

## Current Issue
Valhalla web-app ignores **warnings array** coming from the server, which could potential improve the User experience and help them navigate the issues (clamp errors or any new features errors)

## 👨‍💻 Changes proposed

As stated in the issue #45 there is a requirement to add warnings toast message to the places where server sends warnings.

### Structure of warning (from valhalla package [info.proto](https://github.com/lourduradjou/valhalla/blob/master/proto/info.proto) file)
```
message CodedDescription {
  string description = 1;
  uint64 code = 2;
}

message Info {
  repeated Statistic statistics = 1;      // stats that we collect during request processing
  repeated CodedDescription errors = 2;   // errors that occurred during request processing
  repeated CodedDescription warnings = 3; // warnings that occurred during request processing
  bool is_service = 4;                    // was this a service request/response rather than a direct call to the library
}
```

Going through the docs, the changes were only required at,
- **use-directions-queries.ts**
- **use-isochrones-queries.ts**
- **use-optimized-route-query.ts** 

**Explanation of the changes:**
- An utility file was created to handle this warnings array
- The response data received from the server was checked to have warnings in it, to display the warning message using existing toast from shadcn.

Location: utils/handle-valhalla-warnings.ts
```
  const TOAST_CONFIG: ExternalToast = {
  position: 'bottom-center',
  duration: 5000,
  closeButton: true,
};

export const handleValhallaWarnings = (warnings?: ValhallaWarning[]): void => {
  if (!warnings?.length) return;

  warnings?.forEach((warning) => {
    toast.warning('Routing warning', {
      description: warning.description,
      ...TOAST_CONFIG,
    });
  });
};
```

- **Typescript Types** of the relevant responses where changed to remove the warnings and improve the maintainability and catch bugs.
**Location file: components/types.ts**
Warning was created as a seperate interface to avoid repetition
```
export interface ValhallaWarning {
  code: number;
  description: string;
}
```
**ValhallaIsochroneResponse**, **ValhallaRouteResponse**, **FetchGeocodeObject** in all these interfaces warning structure was added as mentioned in the [docs](https://valhalla.github.io/valhalla/api/turn-by-turn/api-reference/#outputs-of-a-route)

Changed types:
```
export interface ValhallaRouteResponse {
  id: 'valhalla_directions';
  trip: Trip;
  alternates?: ValhallaRouteResponse[];
  warnings?: ValhallaWarning[];
}

export interface ValhallaIsochroneResponse extends GeoJSON.FeatureCollection {
  id: string;
  warnings?: ValhallaWarning[];
}

export interface ValhallaOptimizedRouteResponse {
  trip: Trip;
  id?: string;
  warnings?: ValhallaWarning[];
}
```

## 📄 Note to reviewers

- There are so many places in the docs **Warnings(optional)** were mentioned but I only placed the checking where the data received from the Valhalla server.
- Relevant changes were made directly to the hooks which helps to get the data via axios
- One more place where we could potentially add the check at **nominatim.ts** but it uses **VITE_NOMINATIM_URL=https://nominatim.openstreetmap.org** so considered as not needed.
- I can't simulate the warnings messages via the existing slider (like clamp error), so I changed the data to add warnings in it. (So it proves in future when server returns data with warnings it will work)
- Any endpoints if I missed, please let me know.
---
**New additions: (08/March/2026)**
- Unit test cases were written to check whether it shows warnings when there is a warnings and not when it is empty.
- Existing code was turned into a utils file to remove the redundancy.
- You can find test cases in handle-valhalla-warnings.spec.ts file inside the utils file itself
- Comments were added in the places and also to explain the util function.
- Attached the test cases in the screenshot section.
** (09/March/2026) **
- e2e testing was added for one of the route and test cases were added in the screenshot section.
** (11/March/2026) **
- Required changes in the test cases were added and all relevant files were changed to adapt it.

## Sample output
Valhalla issue warnings

## Test cases outputs
Screenshot from 2026-03-08 20-13-24
Screenshot from 2026-03-08 20-14-33

## e2e test case output
image
image

## AI Applications
- LLM were used to understand the problem and needs as I am new to this repository and ecosystem
- LLM was used to write test cases which is thoroughly checked.
- Code, Description and comments was written by me, I am willing to change even further to help other get an idea of this issue and how I addressed the problem.

### DIFF
--- a/src/components/settings-panel/settings-panel.tsx
+++ b/src/components/settings-panel/settings-panel.tsx
@@ -125,7 +125,7 @@ export const SettingsPanel = () => {
     <Sheet open={settingsPanelOpen} modal={false}>
       <SheetContent
         side="right"
-        className="w-[350px] sm:max-w-[unset] max-h-screen overflow-y-scroll"
+        className="w-[350px] pb-6 sm:max-w-[unset] max-h-screen overflow-y-scroll"
       >
         <SheetHeader className="justify-between">
           <SheetTitle>Settings</SheetTitle>


lourduradjou:
> Yep, first thing is to follow our GSoC contribution guidelines properly, pls update your PR description accordingly.

Hi thank you so much for the feedback, I learnt a lot from this. I can see lot of other contributors are also making the same mistakes like I did and every time you are asking them to review the PR description. Will it be appropriate if I make a PR to add **contributor guidelines file**. And I updated my PR, Looking forward to your review :slightly_smiling_face:

nilsnolde:
cool, thanks.

> Since hooks in this project currently don't have dedicated tests

we have e2e tests. it'd be nice to have at least routing warnings (never mind the other endpoints).

PR #354, opened 2026-03-06
feat: added custom layer editor to tiles panel

### LINKED ISSUE #323: Enable customizing tile viz
### Description

With the verbose tiles, visualization possibilities are almost endless (the biggest hurdle for great viz is probably missing geometry offsetting for singly digitized ways). Would it make sense to be able to provide your own style.json somehow similar to how we handle it for the base map?

### Screenshots

_No response_

### Additional information

_No response_

s4chinjha:
## 🛠️ Fixes Issue

Closes #323

---

## 👨‍💻 Changes proposed

- Added a store to keep track of custom layers the user adds
- Added a dialog with a JSON text area where users can paste a MapLibre layer definition and add it to the map
- The dialog validates the input before adding,checks for valid JSON, a string id field, and that the id is not already on the map
- Custom layers show up in their own section at the bottom of the Tiles panel with a visibility toggle and a remove button
- Custom layers survive map style reloads, they get re-applied automatically
- Custom layers referencing `valhalla-tiles` survive the Valhalla source being toggled off and back on
- Fixed a bug where custom layers were showing up inside standard layer groups as well as the Custom Layers section
- Added accessible labels so screen readers can identify the toggle and remove buttons for each layer
- Added 47 new tests, 836 passing in total

---

## 📷 Screenshots

**Before** (standard tiles)

Screenshot From 2026-03-07 19-11-03

**After** (custom layer added) 

Screenshot From 2026-03-07 19-10-59

Layer definition used above:

```json
{
  "id": "edges-highlight",
  "type": "line",
  "source": "valhalla-tiles",
  "source-layer": "edges",
  "paint": {
    "line-color": "#ff0000",
    "line-width": 2
  }
}
```

This highlights all the edges in red.

---
## 📄 Note to reviewers

- Layers are lost on page refresh, no persistent storage yet

AI Contribution:
I've used AI assistance to help me write most of the code & for debugging the code.
I've reviewed every change, ran all the tests myself, and manually verified the feature works properly.

### DIFF
--- a/.github/workflows/preview-deploy.yml
+++ b/.github/workflows/preview-deploy.yml
@@ -93,7 +93,7 @@ jobs:
       - name: 🌐 Comment preview URL
         uses: actions/github-script@v7
         with:
-          github-token: ${{ secrets.PAT }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
             const prNumber = parseInt('${{ steps.pr.outputs.number }}');
             const { data: comments } = await github.rest.issues.listComments({
--- a/src/components/map/index.tsx
+++ b/src/components/map/index.tsx
@@ -446,10 +446,13 @@ export const MapComponent = () => {
     const dpOpen = state.directionsPanelOpen;
     const spOpen = state.settingsPanelOpen;
 
-    const paddingTopLeft = [screen.width < 550 ? 50 : dpOpen ? 420 : 50, 50];
+    const paddingTopLeft = [
+      window.innerWidth < 550 ? 80 : dpOpen ? 450 : 80,
+      80,
+    ];
     const paddingBottomRight = [
-      screen.width < 550 ? 50 : spOpen ? 420 : 50,
-      50,
+      window.innerWidth < 550 ? 80 : spOpen ? 450 : 80,
+      80,
     ];
 
     mapRef.current.fitBounds(bounds, {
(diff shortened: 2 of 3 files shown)


src/components/tiles/custom-layer-editor.tsx
```diff
@@ -0,0 +1,150 @@
...
+  null,
+  2
+);
+
+export const CustomLayerEditor = () => {
+  const { mainMap } = useMap();
+  const addLayer = useCustomLayersStore((state) => state.addLayer);
+  const layers = useCustomLayersStore((state) => state.layers);
+  const [open, setOpen] = useState(false);
+  const [json, setJson] = useState('');
```
mustaphaturhan: let's call it `value` or `jsonValue`

src/components/tiles/custom-layer-editor.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  const handleOpenChange = (val: boolean) => {
+    setOpen(val);
+    if (!val) {
+      setError(null);
+    }
+  };
+
+  return (
+    <div className="pb-1">
```
mustaphaturhan: I don't think this div is in the correct place. Where we are adding the padding? To the content of dialog, or to the trigger?

s4chinjha:
@nilsnolde @mustaphaturhan here i have tried to implement the changes suggested
In the **first commit** i've renamed json to jsonValue, removed the extra div wrapper and disabled the add Custom Layer button when Append Valhalla layers is off.
And in the **second commit** i've removed the custom-layers-store.

Also i've updated the tests accordingly.

PR #358, opened 2026-03-07
feature: added icon buttons

### LINKED ISSUE #353: few UI improvements: replace text buttons with icon buttons
1. add control buttons for every tab in the upper left corner which, when clicked, shows the left side panel on the appropriate tab. those icon buttons can be ordered vertically, similar to the map controls on the right side but definitely a lot bigger. choose appropriate SVG icons (or generate some).
2. same for the "Open OSM" button: switch it to icon/tool button.

s4chinjha:
## 🛠️ Fixes Issue

Closes #353

## 👨‍💻 Changes proposed

1.Added 3 icon buttons on the left side of the map for Directions, Isochrones and Tiles clicking any of them opens the side panel straight to that tab.

2.Also replaced the "Open OSM" text button in the bottom right with an icon button i.e. same size as the other controls.

3.Built a reusable Tool Button component that all four new buttons use.

4.Added 7 unit tests for Tool Button and also added 4 new tests to make sure the left-side buttons render and navigate correctly and removed the 4 old tests that were testing the same buttons we deleted.

## 📄 Note to reviewers

I've made a Figma dessign for this here -> https://www.figma.com/slides/mK2PwErWWGYkkd8kUZw2i9
I had two design in mind, please give your thoughts on this which you prefer and i can do the changes.
Right now i've implemented the first one.

## 📷 Screenshots

image

AI Contribution:
I've used AI to assist me during this Pr.
The research,design and implementation are part of my work.

### DIFF
--- a/src/components/directions/route-card.spec.tsx
+++ b/src/components/directions/route-card.spec.tsx
@@ -28,6 +28,7 @@ vi.mock('./summary', () => ({
     summary: unknown;
     title: string;
     index: number;
+    routeCoordinates: number[][];
   }) => <div data-testid={`mock-summary-${index}`}>Summary: {title}</div>,
 }));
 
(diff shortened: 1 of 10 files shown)


nilsnolde:
Honestly, I always need code and the live app before I can judge most code/UX proposals. That's what I'd ask you to do usually: no text describing solutions to problems, just code to make it reviewable. Unless it's a big single decision that would alter the entire path of dev of course, it'd be frustrating to develop for hours only to be told to do it entirely differently.

PR #368, opened 2026-03-10
Improve keyboard navigation for profile selector

Shyam-123pandey:
Currently only the active profile button can be focused using Tab navigation.

Closes: #353 

**Telling very hosnestly, u can see above 23 line of code can do such improvement and i saw previous PRs change 1000 line of code just for few improvment (that showing all thing using AI) But In my case All  of You can see How much effort i am providing not to use AI Blindly**

This PR improves accessibility by allowing users to switch between transport
profiles using ArrowLeft and ArrowRight keys.

Changes:
- Added keyboard navigation handler
- Allows cycling through profile buttons with arrow keys
- Improves accessibility for keyboard users

Vedio: 

https://github.com/user-attachments/assets/ae9f55b3-ea37-493a-a5e5-e37318379e5a

### DIFF
--- a/src/components/profile-picker.tsx
+++ b/src/components/profile-picker.tsx
@@ -58,6 +58,28 @@ export const ProfilePicker = ({
     { value: 'motorcycle', label: 'Motorcycle' },
   ];
 
+  const handleKeyDown = (event: React.KeyboardEvent, index: number) => {
+    if (event.key === 'ArrowRight') {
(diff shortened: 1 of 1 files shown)


nilsnolde:
if you're feeling mistreated you can try contacting the gsoc admins and complain, that's your right. personally I've said all I can say.

PR #383, opened 2026-03-18
fix: disable reset buttons for edge cases

### LINKED ISSUE #382: BUG: Shuffling the waypoints enables the delete and reset waypoint buttons, though no location is added in the waypoints.
Shuffling the waypoints enables the delete and reset waypoint buttons, though no location is added in the waypoints.

https://github.com/user-attachments/assets/b132b315-009c-47fa-8e45-24304bf281ca

rohhann12:
…less of order

Closes #382
## 🛠️ Fixes Issue
Minor ux Bug: Shuffling the waypoints enables the delete and reset waypoint buttons, though no location is added in the waypoints.

Changed the logic of how the disabling those buttons were working

## 👨‍💻 Changes proposed
instead of comparing waypoints against defaultWaypoints via JSON.stringify (which breaks on reorder), we simply check if every waypoint's userInput is empty.

In both the files-
- src/components/directions/directions.tsx
-src/components/directions/waypoints/waypoint-item.tsx
I changed the logic of disabling the button instead comparing the waypoints against the default , now we simply check if they are empty.

## 📷 Screenshots

https://github.com/user-attachments/assets/a187e281-41f9-4c73-9e0a-3c7ffb23eb15

### DIFF
--- a/e2e/map.spec.ts
+++ b/e2e/map.spec.ts
@@ -607,8 +607,13 @@ https: test.describe('Left drawer', () => {
 
     await expect(page.getByRole('button', { name: '3' }).first()).toBeVisible();
 
-    // Remove waypoint using the individual remove button (reset is disabled when all inputs are empty)
-    await page.getByTestId('remove-waypoint-button').first().click();
+    // Type something in a waypoint so the reset button becomes enabled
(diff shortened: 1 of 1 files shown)


rohhann12:
mistake- pushed a wrong commit here , will clean it up

PR #385, opened 2026-03-18
feat:add map language control

### LINKED ISSUE #384: i18n: Show English names for country labels on the map
Currently, country names on the map are displayed in their native/local script. For English-speaking users, these should display English names instead.
                                                                                                                                                
This is a map tile/style level change — the tile source or style needs to be configured to prefer name:en (or the equivalent language key) over the default name field for label rendering.

Image

@nilsnolde Let me know if this something you would be interested in being implemented?

Would love to pick this up

rohhann12:
Starting Step to close #384

## 🛠️ Fixes Issue

The issue was the map would show the names of the countries in the format of their own local language (eg- middle eastern countries name would be shown in arabic).

## 👨‍💻 Changes proposed

-src/components/map/index.tsx
-src/components/map/map-language-control.tsx

## 📄 Note to reviewers
Approach:                                                                                                                         
The Shortbread OSM vector tiles use text-field: "{name}" on all symbol layers for label rendering. The tiles also provide name_en and name_de as alternative name properties.
I added a MapLanguageControl component that:      
1. Iterates over all symbol layers in the current map style
2. Swaps the text-field layout property between {name}, {name_en}, or {name_de} based on user selection
3. Persists the choice in localStorage so it survives page reloads and style switches (restored via onLoad and onStyleData events)
Currently limited to 3 languages since that's all Shortbread tiles support. This can be extended once we move to Maptiler tiles which offer broader language coverage. 

## 📷 Screenshots

https://github.com/user-attachments/assets/6194e57b-fb70-447b-bcfa-199c40652e40

### DIFF
--- a/src/components/directions/directions.spec.tsx
+++ b/src/components/directions/directions.spec.tsx
@@ -29,7 +29,7 @@ const mockResults = {
     trip: { summary: unknown; legs: unknown[] };
     alternates?: unknown[];
   },
-  show: { '-1': true },
+  show: { '0': true },
 };
 
 const mockDateTime = { type: 0, value: '2024-01-01T12:00' };
@@ -203,7 +203,7 @@ describe('DirectionsControl', () => {
 
   it('should not render RouteCard when no results', () => {
     render(<DirectionsControl />);
-    expect(screen.queryByTestId('mock-route-card--1')).not.toBeInTheDocument();
+    expect(screen.queryByTestId('mock-route-card-0')).not.toBeInTheDocument();
     expect(screen.queryByText('Directions')).not.toBeInTheDocument();
   });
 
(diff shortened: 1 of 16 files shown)


nilsnolde:
ah now it worked and my eyes didn't see the new map control. I think that's because that's a really strange place to put smth like that. we don't want 2 places with the same intention, language control. we want to rename the "directions language" dropdown and just always show it. then match to german/english for the map, everything else becomes "local".

rohhann12:
hey @nilsnolde 
updated as requested 
1.Removed the separate map language control — unified it with the existing language dropdown in the settings panel
2.Renamed "Directions Language" to "Language" and made it visible on all tabs (not just directions)
3.When language is changed, it now updates both navigation instructions and map labels — en-* maps to English labels, de-DE to German, everything else falls back to local names
4. Added support for both Shortbread (string-based text-field) and Alidade Smooth (expression-based text-field) label formats
5.Alidade Smooth supports language via style modification (not request headers) — the code now handles both tile sources 
https://github.com/user-attachments/assets/32de5f2d-6139-4be7-8b8d-a8f9c5aaa6fd

PR #387, opened 2026-03-18
feat: sync active route selection between map and sidebar

### LINKED ISSUE #369: Improve alternates UX
When requesting with alternative routes, there's hardly any way to correlate the route cards in the left side panel to the alternatives shown on the map.

My suggestion would be to do it like Google Maps (it's the only reasonable way IMO, otherwise I wouldn't care):
- show the primary route by default as we currently do
- if one clicks on any alternative in the map _or_ the sidebar it'll be activated both in sidebar _and_ map

befikre:
Fixes #369

## What was the problem
Alternative routes all looked the same on the map: same color, same style, no way to know which line matched which card in the sidebar. Clicking a route line did nothing. Clicking a sidebar card did nothing to the map. They were completely disconnected.

## What I changed
Introduced activeRouteIndex in the directions store as a single shared value. The map and sidebar both read from it and write to it, so they always stay in sync.

- Active route renders in blue (#3b82f6), full opacity
- Inactive alternatives render in light blue (#66a3ff), 0.5 opacity
- The active route is dynamically sorted to always render on top of inactive ones
- Clicking any route line on the map activates it and highlights the matching sidebar card
- Clicking any sidebar card activates the matching map line
- Primary route (index 0) is active by default on load
- Clicking an already-active card does nothing
- New route request resets active index back to 0

## Files changed
- `directions-store.ts` - added activeRouteIndex state + setter
- `route-lines.tsx` - MapLibre data-driven paint expressions and active route sorting
- `route-card.tsx` - active highlight + click handler
- `directions.tsx` - passes index down to each RouteCard
- `highlight-segment.tsx` - updated for 0-based indexing
- `constants.ts` - updated route colors
- `map/index.tsx` - MapLibre layer click event handler
- `*.spec.tsx` - 3 spec files updated to match the new 0-based indexing

## Tests
797/797 passing 

## ScreenShots
Screenshot 2026-03-19 004459
Screenshot 2026-03-19 004512
Screenshot 2026-03-19 103919

---
_This PR is submitted as a pre-application contribution for GSoC 2026 
(Valhalla - Automated Routing QA project)._


nilsnolde:
This is cool, thanks! We should only revert the color change. Light blue for inactive routes is much better than grey IMO. Blue has cartographic significance: hardly any map element has blue colors other than water bodies. Other than maybe the old watercolor style that hardly changes between map styles. However, try stadiamaps' "Alidade Smooth" basemap with your suggestion and see;)

Rest of the changes are good! Is there anything you want to update in the PR description wrt our GSoC guidelines?

PR #389, opened 2026-03-20
feat: add proper maneuver icons

### LINKED ISSUE #378: Add proper maneuver icons
For the guidance description we currently show a generic icon, the same for every kind of maneuver. We should add support for arrow-like maneuver icons like all other routing apps. I'm honestly never working with instructions, but I think the osrm format is more suited for guidance instructions, so should probably change to that (if it's not what we're using already, not sure at this point).

rohhann12:
## 🛠️ Fixes Issue
-Add proper maneuver icons #378

Closes #378 
## 👨‍💻 Changes proposed

Mapped all the types to lucide react icons                                                                                                
I took the types from here-https://valhalla.github.io/valhalla/api/turn-by-turn/api-reference/
Each maneuver step now shows a contextual icon (arrows for turns, flags for destinations, circles for roundabouts, etc.)

## 📄 Note to reviewers
AI Disclosure- Not using AI to generate descriptions or code 

## 📷 Screenshots
Screenshot 2026-03-20 at 7 52 07 PM

### DIFF
--- a/src/components/directions/directions.tsx
+++ b/src/components/directions/directions.tsx
@@ -14,10 +14,7 @@ import { RouteCard } from './route-card';
 import { parseUrlParams } from '@/utils/parse-url-params';
 import { isValidCoordinates } from '@/utils/geom';
 import { useNavigate } from '@tanstack/react-router';
-import {
-  defaultWaypoints,
-  useDirectionsStore,
-} from '@/stores/directions-store';
+import { useDirectionsStore } from '@/stores/directions-store';
 import {
   useDirectionsQuery,
   useReverseGeocodeDirections,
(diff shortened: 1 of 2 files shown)


nilsnolde (approved):
cool, thanks! 

it's hard to find an icon library which has escalator/elevator etc I guess. we can always later alter an existing lucide icon to fit those. this is more than good enough!

however, I wonder if we're doing the right thing in upstream valhalla maneuver generation. I tried routing ped through berlin main train station and didn't get a "enter building" instruction: 

https://valhalla-app-tests.gis-ops.com/389/directions?profile=pedestrian&wps=13.368592142462802%2C52.526662768718666%2C13.37015926436672%2C52.52399309842485&style=carto

PR #393, opened 2026-03-22
Fix geocode and nominatim on map click

### LINKED ISSUE #371: improve time to rendered route
It's always nagged me that the time between creating/updating waypoints or options and seeing the result on the map takes so much longer than it should. Most of the time is eaten up by nominatim (>90%), probably not because it's slow per se but because it's heavily throttled since it's a public server with way more traffic than ours. Do we really need to geocode by default? I think we should geocode if the user entered an address, but not if waypoints were just placed on the map. 

Also, related: it seems that disabling geocoding in the settings does not bypass the nominatim round trip.

KanishJebaMathewM:
Fixes #371 
proof

proof 2

when i tested the app, (in the network tab) the nominatim was being called each time when i click on the map even though the points have already been fixed. I have added a skip geocode option to skip those calls

### DIFF
--- a/e2e/map.spec.ts
+++ b/e2e/map.spec.ts
@@ -1,23 +1,6 @@
 import { test, expect } from '@playwright/test';
-import {
-  BERLIN_COORDINATES,
-  setupHeightMock,
-  setupNominatimMock,
-  setupRouteMock,
-  setupSearchMock,
-  simpleMockNominatimResponse,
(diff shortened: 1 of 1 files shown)


nilsnolde (commented):
Thanks!

What do you think about extending the scope of the PR and just get rid of reverse geocoding entirely?

PR #396, opened 2026-03-22
feat: permalink with costing options

### LINKED ISSUE #335: UX: extend permalink with costing options
We have a couple of useful attributes we use in the internal router for permalinks, like profile. However, to be truly useful to power users is costing options support in the permalink. Currently we always reset any costing option when reloading the page. Code-wise this is probably pretty painful, also to maintain it in the future. It'd be great if there'd be some kind of integration test that'd catch when a new costing option/setting was added and fail if it's not considered in the permalink.

rohhann12:
## 🛠️ Fixes Issue

closes #335 
## 👨‍💻 Changes proposed
-Added a costing query parameter to the URL that stores any non-default costing options as JSON                                              
- Settings are restored from the URL on page load, so they survive refreshes and can be shared via permalink
- URL stays clean when everything is at defaults — the param only appears when something is changed
  

## 📄 Note to reviewers
Only non-default values are sync to keep the URL short. exclude_polygons is intentionally skipped as it can hold large GeoJSON data.

### DIFF
--- a/e2e/costing-permalink.spec.ts
+++ b/e2e/costing-permalink.spec.ts
@@ -1,0 +1,91 @@
+import { test, expect, type Page } from '@playwright/test';
+
+const BASE_URL = 'http://localhost:3000';
+
+async function openSettingsPanel(page: Page) {
+  await page.getByTestId('tab-directions-button').click();
+  await page.getByTestId('show-hide-settings-btn').click();
+  // wait for the settings panel content to be visible
+  await expect(
+    page.getByRole('checkbox', { name: 'Shortest', exact: true })
+  ).toBeVisible();
+}
+
+test.beforeEach(async ({ page }) => {
+  await page.goto(`${BASE_URL}/directions?profile=car`);
+  await openSettingsPanel(page);
+});
+
+test('costing options appear in URL when a setting is changed', async ({
+  page,
+}) => {
+  await page.getByRole('checkbox', { name: 'Shortest', exact: true }).click();
+
+  await expect(page).toHaveURL(/costing=/);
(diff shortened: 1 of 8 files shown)


nilsnolde (changes requested):
thanks! never mind the geojson, it's ok to skip that. 

I just tried it a little bit and I like how it works! I'll let @mustaphaturhan judge the code, doesn't look too crazy to me.

few questions/concerns:
- when we add some (scalar) costing option in the future, will it be auto-supported or smth needs to be done for URL syncing?
- we should add a (few) e2e tests for this too. not too exhaustive, just making sure it won't generally break in the future.

nilsnolde (approved):
cool, this works for me!

rohhann12:
Hi @mustaphaturhan  — just checking in to see if you've had a chance to review this. @nilsnolde has already approved it and all checks are passing. Happy to make any changes you'd like before merging!

rohhann12:
Hi @mustaphaturhan — just a ping on this one! @nilsnolde  has approved and CI is green. Let me know if there's anything you'd like me to change, happy to iterate. Would love to get this wrapped up.

mustaphaturhan (approved):

rohhann12:
Hi @nilsnolde  — thanks so much for the reviews! Both approvals are in and CI is green. Whenever you get a chance, feel free to merge. Happy to rebase if needed too!

rohhann12:
Hi @nilsnolde  — just a gentle reminder on #396  whenever you get a chance! Both approvals are in and CI is green. Happy to rebase if needed. Thanks!

PR #399, opened 2026-03-23
Fix (Waypoints): clear placeholder results on panel close (#365)

### LINKED ISSUE #365: [Bug] Waypoint populated on reopening the left side panel
Waypoint automatically get added on closing and opening the left side panel

Steps to reproduce -

1. Add three waypoint
2. Close the sidebar
3. Open the sidebar  ( you will see a waypoint added ) repeat this you can see multiple waypoints on the left side panel of the direction tab.

Video -

https://github.com/user-attachments/assets/6c1b2c0e-4108-40f0-ba42-7af90999696e

AaravJalan:
## 🛠️ Fixes Issue

Closes #365

When the directions panel was closed and reopened, waypoints that were in a loading/placeholder state (coordinate pinned on map, reverse geocode still in flight, or just completed) would re-render with stale data — making it appear as if a waypoint had been populated when the user hadn't finished setting one.

### `src/stores/directions-store.ts`
- Added `clearPlaceholderWaypoints` action to `DirectionsActions` interface
- Implemented the action: iterates waypoints and clears any that contain only placeholder geocode results (those with empty `title`), resetting both `geocodeResults` and `userInput` to empty

### `src/components/route-planner.tsx`
- Call `clearPlaceholderWaypoints()` before `toggleDirections()` on sheet close, so placeholder state is always cleaned up when the panel is dismissed

### `src/components/directions/waypoints/waypoint-item.tsx`
- Look up waypoint by `id` instead of `index` to prevent stale index references from causing wrong data to render after array mutations

## Type of Change
- [x] Bug fix

### DIFF
--- a/e2e/map.spec.ts
+++ b/e2e/map.spec.ts
@@ -1,23 +1,6 @@
 import { test, expect } from '@playwright/test';
-import {
-  BERLIN_COORDINATES,
-  setupHeightMock,
-  setupNominatimMock,
-  setupRouteMock,
-  setupSearchMock,
-  simpleMockNominatimResponse,
-} from './helpers';
+import { setupHeightMock, setupRouteMock, setupSearchMock } from './helpers';
 import { customRouteResponse } from './mocks';
-
-interface NominatimApiRequest {
-  url: string;
-  method: string;
-  params?: {
-    format?: string | null;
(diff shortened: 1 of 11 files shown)


AaravJalan:
GSOC Contribution

PR #402, opened 2026-03-23
fix(routing): skip nominatim reverse geocode for map-placed waypoints

### LINKED ISSUE #371: improve time to rendered route

AaravJalan:
## 🛠️ Fixes Issue

Closes #371 

The main bottleneck in route render time was the Nominatim reverse geocode calls for waypoints placed via map click. 

### `src/hooks/use-directions-queries.ts`
- Check `use_geocoding` from the store before calling Nominatim in `useReverseGeocodeDirections.` If false, build a coordinate-only `ActiveWaypoint` directly with zero network requests.
- Remove `setTimeout(..., 500)` around `showLoading(false).`
- Wrap `zoomTo` in `requestAnimationFrame` so route layer paints first.

### `src/components/map/parts/map-context-menu.tsx`
- Add optional `lngLat` parameter to `onAddWaypoint` callback.
- Pass `popupLocation` through on all three waypoint buttons so the coordinates from the right-click event flow through to the handler.

### `src/components/map/index.tsx`
- Update `handleAddWaypoint` to accept and use the `lngLat` argument passed from `MapContextMenu`.

## Type of Change
- [x] Performance improvement
- [x] Bug fix

### DIFF
--- a/src/components/map/index.tsx
+++ b/src/components/map/index.tsx
@@ -305,13 +305,15 @@ export const MapComponent = () => {
       });
   }, []);
 
+  // handleAddWaypoint now receives the lngLat from MapContextMenu directly.
   const handleAddWaypoint = useCallback(
-    (index: number) => {
-      if (!popupLngLat) return;
+    (index: number, lngLat?: { lng: number; lat: number }) => {
+      const location = lngLat ?? popupLngLat;
+      if (!location) return;
       setShowContextPopup(false);
 
       updateWaypointPosition({
-        latLng: { lat: popupLngLat.lat, lng: popupLngLat.lng },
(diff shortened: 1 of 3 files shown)


nilsnolde:
closed in #393

AaravJalan:
GSOC Contribution

PR #404, opened 2026-03-24
feat: add access_restrictions MVT layer visualization

### LINKED ISSUE #390: add `access_restrictions` MVT layer
ref https://github.com/valhalla/valhalla/pull/5912

Sherley-Sonali:
## 🛠️ Fixes Issue
Closes #390

## 👨‍💻 Changes proposed
- Added two new layer definitions in `valhalla-layers.ts`:
  - `valhalla-access-restrictions-permanent` - solid red line for permanent restrictions
  - `valhalla-access-restrictions-timed` - dashed orange line for timed/conditional restrictions
- Updated `index.tsx` to include both layers in click detection and interactive layer IDs
- Updated `valhalla-layers.spec.ts` and `valhalla-layers-toggle.spec.tsx` with tests for the new layers

## 📄 Note to reviewers
Styled permanent and timed restrictions differently rather than uniformly since they behave differently at runtime and a debugging tool should make that distinction clear at a glance. Used AccessType enum values (6, 7) from the backend to filter timed restrictions. Able to adjust if uniform styling is preferred.

## 📷 Screenshots
The access_restrictions layer renders two visually distinct styles:
- Solid red lines for permanent restrictions
- Dashed orange lines for timed/conditional restrictions

Toggling the layer on/off from the tiles panel shows/hides both.
Screenshot from 2026-03-24 12-57-11
Screenshot from 2026-03-24 12-57-18

### DIFF
--- a/src/components/map/index.tsx
+++ b/src/components/map/index.tsx
@@ -50,6 +50,8 @@ import {
   VALHALLA_EDGES_LAYER_ID,
   VALHALLA_NODES_LAYER_ID,
   VALHALLA_SHORTCUTS_LAYER_ID,
+  VALHALLA_ACCESS_RESTRICTIONS_PERMANENT_LAYER_ID,
+  VALHALLA_ACCESS_RESTRICTIONS_TIMED_LAYER_ID,
 } from '@/components/tiles/valhalla-layers';
 import { MarkerIcon, type MarkerColor } from './parts/marker-icon';
 import { maxBounds } from './constants';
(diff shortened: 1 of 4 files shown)


nilsnolde (commented):
thanks! IMO the high-level classification is fine. 

a few things:
- red/orange is already too present IMO. but not sure what else makes sense;) maybe it's fine.. 
- definitely we want to adapt the popup to show proper strings for `type`  & `modes`. `value` should also have a unit depending on `type`. 

the latter can also be done in a follow-up PR. wdyt @chrstnbwnkl ?

chrstnbwnkl (approved):
The default color scheme requires some general overhaul if we're planning to show more layers like this, so this can be addressed separately.

Sherley-Sonali:
I'll pick up the popup improvements as a follow-up PR.

On the color scheme: muting edges/nodes makes sense. One thing that reinforces it structurally: layer z-order. Restrictions defined last in the layer stack are never occluded regardless of how color calibration lands across different base layers. So, visual hierarchy and render order working together makes it more robust than color alone.

PR #405, opened 2026-03-27
feat: add elevation support to export

### LINKED ISSUE #211: Add option to include elevation for download
### Description

Missing elevation when downloading json

### Screenshots

_No response_

### Additional information

_No response_

a-y-a-n-das:
## 🛠️ Fixes Issue

feat: Add option to include elevation for download

Closes #211 

## 👨‍💻 Changes proposed
- added a prop in `buildDirectionRequest` to include `elevations`.
- updated `useDirectionQuery` to accept a prop for the same and return directions with `elevations`.
- added a checkbox for elevation in the dropdown.
- With checked, it makes an additional call with params: `elevation_interval: 30` before downloading.

## 📄 Note to reviewers
AI: used for generating tests.

## 📷 Screenshots
image

### DIFF
--- a/src/components/directions/route-card.spec.tsx
+++ b/src/components/directions/route-card.spec.tsx
@@ -6,6 +6,8 @@ import type { ParsedDirectionsGeometry } from '@/components/types';
 
 const mockExportDataAsJson = vi.fn();
 const mockDownloadFile = vi.fn();
+const mockFetchHeight = vi.fn();
+const mockToastError = vi.fn();
 
 vi.mock('@/utils/export', () => ({
   exportDataAsJson: (...args: unknown[]) => mockExportDataAsJson(...args),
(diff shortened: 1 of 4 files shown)


nilsnolde (changes requested):
A bunch of things are not good tbh:

- the UI is not up to my standards, looks & feels really odd
image

- this PR breaks the routing logic: if the page is freshly reloaded and we add 2 waypoints in the map it won't auto-request the route, which it needs to do obviously
- doing another /route request is way too heavy! I know this feature will hardly be ever used by people, so in practice it doesn't matter much. it's just a good principle to not waste any byte & CPU cycle. use /height instead, that's pretty much for free.

a-y-a-n-das:
Yes, sirrr
- A checkbox without any border looks odd.
- I tried reloading the preview, and it's fetching the routes.
- I actually thought of using something other than `/route` (like the height graph data).

I will try to use that and touch up the UI.

a-y-a-n-das:
image

Now using `/height`
Implemented using `fetch` due to this #413

a-y-a-n-das:
Please let me know if any further changes are needed.

PR #412, opened 2026-03-31
Add expansion tile visualization

tejash-sr:
- Add a reusable source-layer toggle for vector tile sources.
- Add expansion layer helpers and a visible expansion debug layer.
- Wire the map and tiles panel to query and display expansion features.
- Update popup labeling and tests for the new expansion flow.

### DIFF
--- a/src/components/map/index.spec.tsx
+++ b/src/components/map/index.spec.tsx
@@ -823,6 +823,46 @@ describe('MapComponent', () => {
       vi.useRealTimers();
     });
 
+    it('should query expansion layers when they exist', async () => {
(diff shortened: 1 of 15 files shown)


chrstnbwnkl:
A couple of things: 

  - please reference the issue this is resolving 
  - add screenshots that show what this PR accomplishes, and
  - most importantly, Valhalla does not serve "expansion tiles", the end point you are querying does not output MVT 
 
Which makes me think you didn't even try this out locally (you would just get a bunch of 400 error responses from Valhalla complaining about invalid parameters for `/expansion`).

PR #414, opened 2026-04-26
add x-client-id header

nilsnolde:
## 👨‍💻 Changes proposed

This PR sends the `X-Client-Id` header to the valhalla service, so we can tell how many requests are done through the public web app (if they used the public service).


chrstnbwnkl:
Would it make sense to make this a `.env` var and use some non-default value for our deployment? Just to differentiate between people running a fork or a locally hosted version. Shouldn't be the case much but it's low hanging fruit

nilsnolde:
yep, added in https://github.com/valhalla/web-app/commit/4ae8d1a5e9e4c635086f37021064e9ebdc804630

PR #418, opened 2026-05-09
feat: more intuitive colors for waypoint markers

nilsnolde:
origin: green
destination: red
via: grey


nilsnolde:
I'll keep improving a few things. never mind that it'll fail deployment until the valhalla servers are fully back online. I'll manually deploy the web app at some point tonight.

